PR #37, opened 2025-06-20
Fix multi/parallel-tool invocation

Evyatar108:
The router fails with "Invalid Parameters" when more than one tool is triggered in an API call to the LLM, this PR fixes it


musistudio:
Thanks for your contribution!

PR #79, opened 2025-06-30
fix(server): handle exceptions in messages route

screw-hand:
This PR includes two small but important fixes to improve the overall usability and robustness of the `claude-code-router`.

**Changes:**

1.  **Corrected Outdated Command in Status Message (`src/utils/status.ts`)**
    * The status message displayed by `ccr status` was suggesting the outdated command `ccr close`.
    * This has been updated to the correct command, `ccr stop`, to avoid user confusion and align with the current CLI implementation.

2.  **Added Exception Handling for `/v1/messages` Route (`src/index.ts`)**
    * Previously, if a downstream API call (e.g., to OpenAI/DeepSeek) failed, the server would catch the exception but not send a response to the client. This caused the client's request to hang until it timed out.
    * An error handling mechanism has been added to the `catch` block. Now, if an error occurs, the server will immediately respond with a `500 Internal Server Error` status and a JSON body containing the error details. This ensures the client receives a prompt and clear response upon failure.

These changes enhance the user experience and make the server's behavior more predictable and reliable.

Thank you for your review and consideration!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -153,6 +153,7 @@ async function run(options: RunOptions = {}) {
       await streamOpenAIResponse(res, completion, req.body.model, req.body);
     } catch (e) {
       log("Error in OpenAI API call:", e);
+      return res.status(500).json({ error: e });
     }
   });
   server.start();
--- a/src/utils/status.ts
+++ b/src/utils/status.ts
@@ -2,10 +2,10 @@ import { getServiceInfo } from './processCheck';
 
 export function showStatus() {
     const info = getServiceInfo();
-    
+
     console.log('\n📊 Claude Code Router Status');
     console.log('═'.repeat(40));
-    
+
     if (info.running) {
         console.log('✅ Status: Running');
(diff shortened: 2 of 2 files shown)


GinWU05:
Closing this pull request as the proposed changes have been superseded by a recent, significant refactoring in the upstream branch.

The author's commit 30c764828a3556475e86cc073ac122bd307f0baf has addressed the two issues this PR intended to fix:

1.  **Outdated Command in Status Message (`src/utils/status.ts`)**: This issue was resolved directly by the author during the manual refactoring within the aforementioned commit.
2.  **Exception Handling for `/v1/messages` Route (`src/index.ts`)**: This fix is no longer applicable. The refactoring introduced the `@musistudio/llms` package, which replaces the underlying logic for this route. The original problem space has been eliminated by this new dependency.

Given that both objectives of this PR have been accomplished or rendered obsolete by the upstream changes, this contribution is no longer necessary.

PR #444, opened 2025-08-05
feat: non interactive spawn compatible

jagnani73:
**Purpose:** Enables compatibility with non-interactive environments like GitHub Actions, Docker containers, and CI/CD systems
**Functionality**: Sets appropriate environment variables (CI=true, FORCE_COLOR=0, etc.) and configures stdin handling
**Use Case**: Prevents processes from hanging in automated environments
**Example:** Shows how to use it in GitHub Actions workflows
**Default Value**: false to maintain backwards compatibility


musistudio:
In a CI environment, doesn't ccr code pass all current environment variables to the claude command?

jagnani73:
It does in a CI environment
I am trying it in a webcontainer environment and there I am getting the error

```
Error: Raw mode is not supported on the current process.stdin, which Ink uses as input stream by default.
Read about how to prevent this error on https://github.com/vadimdemedes/ink/#israwmodesupported
```

This is likely because of the `inherit` property passed in the `spawn`

jagnani73:
> In a CI environment, doesn't ccr code pass all current environment variables to the claude command?

It does in a CI environment
I am trying it in a webcontainer environment and there I am getting the error

```
Error: Raw mode is not supported on the current process.stdin, which Ink uses as input stream by default.
Read about how to prevent this error on https://github.com/vadimdemedes/ink/#israwmodesupported
```

This is likely because of the `inherit` property passed in the `spawn`

musistudio:
> webcontainer

Can WebContainer run Claude Code? I tried it before and ran into a lot of issues.

jagnani73:
> > webcontainer
> 
> Can WebContainer run Claude Code? I tried it before and ran into a lot of issues.

Yes, I have it working
image

jagnani73:
Thank you for accepting the PR @musistudio 
When are you creating a new release with this?

musistudio:
I released v1.0.33

PR #592, opened 2025-08-15
Add tool use routing mechanism and update configuration

trainsh:
- Introduced a new "toolUse" configuration in config.example.json to support tool/function calls.
- Enhanced the router logic in router.ts to handle tool use routing, including checks for ongoing tool conversations.
- Updated README files to document the new "toolUse" feature.
- Added corresponding UI components and localization for "toolUse" in both English and Chinese.
- Updated package-lock.json with new dependencies for the UI components.


listeng:
Poe now fully supports tool usage.

PR #607, opened 2025-08-17
feat: global model router

### LINKED ISSUE #606: [Feature Request] UI global model
In the UI, switching the model requires switching a series of models.
Is it possible to add a global component to select all types of models?

Sn0wo2:
fix #606

### DIFF
--- a/ui/src/components/Router.tsx
+++ b/ui/src/components/Router.tsx
@@ -40,6 +40,20 @@ export function Router() {
     setConfig({ ...config, Router: newRouter });
   };
 
+  const handleGlobalModelChange = (value: string) => {
+    if (!value) return;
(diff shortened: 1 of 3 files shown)


geocine:
isn't `default` the global router?

Sn0wo2:
> isn't `default` the global router?

didnt test it, but isnt `Default` just the **fallback**? If not, then whats the point of the other types?

geocine:
Base on my observation it always starts with `default` and switches to others based on whats happening. 

If it tries to run a background task it will use `background`. If you switch to plan mode it switches to `thinking`. When it hits the context threshlod it switches to `long context`. When it need to request a `websearch` it switches to that.

PR #836, opened 2025-09-24
Reset CLAUDE_CODE_USE_BEDROCK when running with ccr code.

RobinChiu:
If CLAUDE_CODE_USE_BEDROCK is set in env, there will be an error. Need to unset it.


musistudio:
Thank you for your contribution!

PR #876, opened 2025-10-05
Feature/add model selector command

TonyGeez:
### Description
Add a new CLI command `ccr model` that provides an interactive interface for managing models and providers from the config.json.

### Features
- Update models: Select from available models for any configuration type
- Add new models: Add models to existing providers
- Add new providers: Add new provider
- Transformer: Can set basic transformer when create new provider

### Changes
Added src/utils/modelSelector.ts
Modified src/cli.ts (adding model command handler
Added @inquirer/prompts to package.json


TonyGeez:
Closing this PR, will create a new one with updated fixes.

PR #879, opened 2025-10-06
Fix boolean flag handling for Claude Code options

2008sliu:
Problem:
When passing boolean flags like --continue and --resume to Claude Code through ccr, they were incorrectly transformed to "--continue true" and "--resume true". Claude Code expects these as standalone flags without values, causing them to not be recognized.

Solution:
Modified the argument reconstruction logic in codeCommand.ts to check if a flag value is boolean true. Boolean flags are now passed without values (e.g., "--continue"), while non-boolean flags retain their values (e.g., "--option value").


musistudio:
Thank you for your contribution!

2008sliu:
You are awesome to respond so soon.
This is an awesome project which benefitted me a lot.
Thanks

jarlbrak:
@2008sliu I am seeing odd behavior from --resume where it is re-loading the CLAUDE.md context again. Are you seeing this as well?

PR #1128, opened 2026-01-07
Examples + documentation

saidelike:
- translate examples/README.md from Chinese to English
- review main README.md titles and hierarchy + link to examples/READMEs
- add .json config example showing how to use OpenAI models with Claude Code


saidelike:
NOTE for reviewer: I advise to read every single commit one at a time (instead of looking at all the changes in once). This is because I moved files and so it will be way faster to review this way.

PR #1210, opened 2026-02-12
feat: add Qwen CLI transformer support and fix custom transformer loa…

thrun-11:
…ding

This commit adds built-in support for Qwen Code CLI as a transformer and fixes two critical bugs in the transformer system:

**New Features:**
- Add QwenCLITransformer as a built-in transformer for Qwen Code CLI integration
- Supports OAuth-based authentication with automatic token refresh
- Forces 'coder-model' for all requests to ensure compatibility
- Adds proper User-Agent header for Qwen CLI API

**Bug Fixes:**
1. Custom transformer loading (transformer.ts):
   - Previously, code tried to instantiate module.exports directly with 'new module()'
   - This failed because require() returns an exports object, not the class
   - Now properly handles CommonJS exports (module.exports), ES module default exports, and named exports by detecting the actual class constructor

2. Model-specific transformer result handling (routes.ts):
   - Previously, model-specific transformers assigned the entire return value to requestBody
   - This caused the wrapper object {body, config} to be sent instead of just body
   - Now extracts body and merges config, consistent with provider-level transformers

**Configuration:**
Users can now use Qwen CLI by configuring a provider with the 'qwen-cli' transformer:

  {
    name: qwen,
    api_base_url: https://portal.qwen.ai/v1/chat/completions,
    api_key: dummy,
    models: [coder-model],
    transformer: {
      use: [qwen-cli]
    }
  }

OAuth credentials are automatically managed from ~/.qwen/oauth_creds.json

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1139 @@
+{
+  "name": "@musistudio/claude-code-router",
+  "version": "2.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@musistudio/claude-code-router",
+      "version": "2.0.0",
+      "license": "MIT",
+      "bin": {
+        "ccr": "dist/cli.js"
+      },
+      "devDependencies": {
+        "@types/node": "^24.0.15",
+        "esbuild": "^0.25.1",
+        "shx": "^0.4.0",
+        "typescript": "^5.8.2"
+      },
+      "engines": {
+        "node": ">=20.0.0",
+        "pnpm": ">=8.0.0"
+      }
+    },
+    "node_modules/@esbuild/aix-ppc64": {
+      "version": "0.25.12",
+      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.25.12.tgz",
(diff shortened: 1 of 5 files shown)


jeugregg:
do you think with your PR, ccr will be fixed about quota limit message ? 
Because, after using official ccr with qwen, after a while,  if I use qwen directly, I still have enough free tokens but not with ccr  : error:  `429 {"error":{"message":"Error from provider(qwen,qwen3-coder-plus: 429): {\"error\":{\"code\":\"insufficient_quota\",\"message\":\"Free allocated quota
exceeded.`
How to test your version ? I am not sure to understand how to install dev tools. maybe someone have a link from the doc ?

thrun-11:
@jeugregg Yes, the quota limit message will be fixed in my PR. I’m currently using CCR with Qwen. Qwen code has introduced restrictions on third-party usage of Qwen, such as kilocode and CCR with Qwen, which is why the error message appears. To address this issue, I’ve added a Qwen transformer that mimics the request sent from CCR to the Qwen server, ensuring it matches the official request sent from Qwen code cli to the Qwen server.
you can clone this and build it. before using ccr first run qwen to authenticate your oauth token then set the config and start using it.

PR #1211, opened 2026-02-13
fix(cli): avoid DEP0190 in code command spawn

### LINKED ISSUE #974: DeprecationWarning
When I run ccr code I get the message
`(node:433077) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
(Use `node --trace-deprecation ...` to show where the warning was created)`

AbdelrahmanHafez:
Fixes https://github.com/musistudio/claude-code-router/issues/974
`ccr code` currently calls `spawn(command, args, { shell: true })`, which triggers Node's DEP0190 warning because args are concatenated by the shell.

This change keeps shell behavior, but switches to a single safely quoted command string and calls `spawn(commandString, { shell: true })`. Argument forwarding stays intact, including values with spaces.

## Test Plan
- [x] `pnpm build`
- [x] `CLAUDE_PATH=echo NODE_OPTIONS=--trace-deprecation node dist/cli.js code --help`
- [x] `CLAUDE_PATH=echo node dist/cli.js code --model "claude sonnet" --verbose`

### DIFF
--- a/packages/cli/src/utils/codeCommand.ts
+++ b/packages/cli/src/utils/codeCommand.ts
@@ -5,9 +5,8 @@ import {
   incrementReferenceCount,
   closeService,
 } from "./processCheck";
-import { quote } from 'shell-quote';
-import minimist from "minimist";
 import { createEnvVariables } from "./createEnvVariables";
+import { quote } from "shell-quote";
 
 export interface PresetConfig {
   noServer?: boolean;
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez:
Closing, the repo was rewritten and `packages/cli` no longer exists, so this fix no longer applies.

PR #1220, opened 2026-02-19
Fix: use CLAUDE_CODE_ATTRIBUTION_HEADER to avoid breaking cache

timhok:
Claude Code version >= 2.1.36 now sending special header that changes in every request, it looks like this:
`text:"x-anthropic-billing-header: cc_version=xxxx; cc_entrypoint=cli; cch=xxxx;"`
This breaks prompt cache and triggers full prompt processing on every request, slowing down local inference significantly.
When CLAUDE_CODE_ATTRIBUTION_HEADER is set to 0 in env, claude code stops sending this header and prompt cache works.

Undocumented variable is taken from https://github.com/anthropics/claude-code/issues/24168

### DIFF
--- a/packages/cli/src/types.d.ts
+++ b/packages/cli/src/types.d.ts
@@ -70,6 +70,7 @@ interface ClaudeSettingsFlag {
     NO_PROXY: string;
     DISABLE_TELEMETRY: string;
     DISABLE_COST_WARNINGS: string;
+    CLAUDE_CODE_ATTRIBUTION_HEADER: any;
     API_TIMEOUT_MS: string;
     CLAUDE_CODE_USE_BEDROCK?: undefined;
     [key: string]: any;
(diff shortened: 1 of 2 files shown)


phpmac:
Great fix. Worth flagging why this matters beyond Bedrock's 400 error:

The `x-anthropic-billing-header` string isn't sent as an HTTP header — Claude Code >= 2.1.36 injects it as a **text block at the front of the system prompt** (per the binary analysis in anthropics/claude-code#24168). Because it sits in the prompt prefix and its `cch=` segment changes per request, the prompt-cache prefix breaks on every turn → full reprocessing of the entire context each time.

This silently kills prompt cache for **any non-official base URL** (self-hosted inference, third-party gateways, local models routed through CCR), not just Bedrock. The impact scales with context length — long sessions get progressively slower because the whole history gets re-prefilled every turn.

Defaulting `CLAUDE_CODE_ATTRIBUTION_HEADER=0` is the right call since CCR already redirects traffic away from the official endpoint, so the attribution string serves no purpose downstream.

Minor thought: if enterprises ever rely on that header for internal billing/attribution, exposing it via `claudeCodeSettings.env` (overridable per-preset) instead of purely hardcoding would be more flexible — but as a default, `"0"` is correct.

PR #1285, opened 2026-03-21
fix: strip unsupported thinking field from responses input

KoSukeWork:
## Problem

When using the OpenAI Responses transformer, some messages may carry a `thinking` field from prior chat history.

This field is not accepted by OpenAI-compatible Responses providers and can cause errors such as:

`Unknown parameter: 'input[72].thinking'`

## Fix

Sanitize message objects before pushing them into `input[]` and remove unsupported fields:

- `thinking`
- `cache_control`

## Scope

This PR only changes request sanitization in:

- `packages/core/src/transformer/openai.responses.transformer.ts`

No CLI, tool schema, or response conversion logic is changed.

### DIFF
--- a/packages/core/src/transformer/openai.responses.transformer.ts
+++ b/packages/core/src/transformer/openai.responses.transformer.ts
@@ -110,26 +110,31 @@ export class OpenAIResponsesTransformer implements Transformer {
     request.messages.forEach((message) => {
       if (message.role === "system") return;
 
-      if (Array.isArray(message.content)) {
-        const convertedContent = message.content
-          .map((content) => this.normalizeRequestContent(content, message.role))
+      const sanitizedMessage: any = { ...message };
+      delete sanitizedMessage.thinking;
+      delete sanitizedMessage.cache_control;
+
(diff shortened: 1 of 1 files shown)


KoSukeWork:
This PR fixes the same class of error as #1266, but does the cleanup at the point where messages are converted into Responses `input[]`, instead of mutating `request.messages` globally.

Both patches are addressing the same compatibility issue: some strict OpenAI Responses providers reject `thinking` fields carried over from message history.

The main difference in my PR is where the sanitization happens:

- #1266 mutates `request.messages` in place before transformation
- this PR sanitizes each message right before it is pushed into the Responses `input[]`

I chose that approach because it keeps the cleanup scoped to the OpenAI Responses request path and avoids mutating the original `request.messages` object in place.

So in short:

- same problem being fixed
- same goal: avoid strict-provider 400s
- different implementation tradeoff:
  - #1266 is smaller
  - this PR is a bit more localized / lower-side-effect

If maintainers prefer, I am happy to align with the style of #1266 or rework this PR toward a common approach.

PR #1320, opened 2026-04-07
fix(vertex-gemini): add ?alt=sse for streaming endpoint

### LINKED ISSUE #1315: Bug: vertex-gemini transformer missing ?alt=sse for streaming requests
## Bug Description

The `vertex-gemini` transformer in CCR v2.0.0 crashes on all streaming requests with a 500 error:

```
TypeError: Cannot read properties of undefined (reading '0')
    at mD (cli.js:647:7645)
```

Non-streaming requests work fine. Streaming requests fail 100% of the time.

## Root Cause

The `gemini` transformer and `vertex-gemini` transformer build streaming URLs differently:

```javascript
// gemini transformer (correct):
url: `./${model}:${stream ? "streamGenerateContent?alt=sse" : "generateContent"}`

// vertex-gemini transformer (missing ?alt=sse):
url: `./v1beta1/projects/${project}/locations/${location}/publishers/google/models/${model}:${stream ? "streamGenerateContent" : "generateContent"}`
```

Without `?alt=sse`, Vertex AI's `streamGenerateContent` endpoint returns a JSON array format (`[{...},{...}]`) instead of SSE format (`data: {...}`). The CCR streaming parser (`mD` function) expects SSE format, so it fails when trying to access `candidates[0]` on the unexpected response structure.

Adding `?alt=sse` makes Vertex AI return standard SSE format, which resolves the issue.

## Suggested Fix

In the `IN` class (vertex-gemini transformer) `transformRequestIn` method, change `streamGenerateContent` to `streamGenerateContent?alt=sse`. This is a one-line fix that aligns `vertex-gemini` with the existing `gemini` transformer behavior.

## Impact

All users of the `vertex-gemini` transformer with streaming enabled (which is the default for Claude Code) are affected.

soeric:
Fixes #1315

Without `?alt=sse`, Vertex AI's `streamGenerateContent` endpoint returns JSON array format instead of SSE format, causing CCR's streaming parser to fail with a 500 error (TypeError: Cannot read properties of undefined (reading '0')).

One-line fix: append `?alt=sse` to the streaming URL in `vertex-gemini.transformer.ts`, consistent with the existing `gemini.transformer.ts` implementation.

Credit: @Matsubarai independently identified and fixed this issue in their fork (Matsubarai/claude-code-router#1).

### DIFF
--- a/packages/core/src/transformer/vertex-gemini.transformer.ts
+++ b/packages/core/src/transformer/vertex-gemini.transformer.ts
@@ -58,7 +58,7 @@ export class VertexGeminiTransformer implements Transformer {
       body: buildRequestBody(request),
       config: {
         url: new URL(
-          `./v1beta1/projects/${projectId}/locations/${location}/publishers/google/models/${request.model}:${request.stream ? "streamGenerateContent" : "generateContent"}`,
+          `./v1beta1/projects/${projectId}/locations/${location}/publishers/google/models/${request.model}:${request.stream ? "streamGenerateContent?alt=sse" : "generateContent"}`,
             provider.baseUrl.endsWith('/') ? provider.baseUrl : provider.baseUrl + '/' || `https://${location}-aiplatform.googleapis.com`
         ),
         headers: {


Matsubarai:
Hey @soeric ! I am working on it and face a new issue. In `ccr code`, I try to chat after this fix, but there is no response and no error reported in the chat session (with some model thinking delay). Have you ever met with this issue?

soeric:
Hi @Matsubarai! I haven't encountered this issue on my end so far. But I recently noticed a similar symptom (thinking delay with no response) when using GPT-5.4 Pro through the chat completions API — it turned out to be a provider-side issue with that specific endpoint, not related to CCR itself.

Your issue might be something similar — could you share more details?

Which model/provider are you using?
What do the CCR logs show?
It's possible this is an upstream API issue rather than something introduced by the ?alt=sse fix.

PR #1332, opened 2026-04-15
fix: DEP0190 DeprecationWarning from exec() with shell in CLI

### LINKED ISSUE #1324: [Bug] DEP0190 DeprecationWarning in CLI when using shell option with exec/spawn
## Description

When running `ccr code` or other commands that start the service, a Node.js deprecation warning is displayed:

```
(node:24541) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
```

## Root Cause

The issue is in `packages/cli/src/cli.ts` at line 269 where `exec()` is used with a string command:

```typescript
exec(openCommand, (error) => {
  // ...
});
```

The `exec()` function internally uses `shell: true`, which triggers this deprecation warning in newer Node.js versions.

Additionally, there may be similar issues with `spawn()` calls that could benefit from the same security improvements.

## Security Implications

While the current usage appears to use hardcoded strings for platform detection (`open`, `start`, `xdg-open`), this pattern could become vulnerable if user input is ever incorporated into these commands without proper escaping.

## Proposed Solution

Replace the `exec(openCommand, ...)` call with a safer approach:

### Option 1: Use `spawn` without shell (Recommended)
```typescript
// Instead of: exec(openCommand, callback)
// Use:
const [command, ...args] = openCommand.split(' ');
const openProcess = spawn(command, args);
openProcess.on('error', (error) => {
  console.error("Failed to open browser:", error.message);
  process.exit(1);
});
```

### Option 2: If shell is absolutely required, properly escape arguments
```typescript
import { quote } from 'shell-quote';

const safeUrl = quote([uiUrl]);
// Then construct the command with safeUrl
```

## Additional Recommendations

1. Review all `spawn` and `exec` calls in the codebase for similar patterns
2. Consider adding a linter rule to prevent unsafe child_process usage
3. Add proper error handling for the browser opening functionality

## Environment
- Node.js version: >=18 (where this deprecation warning appears)
- ccr version: 2.0.0

This fix will eliminate the deprecation warning and improve the security posture of the CLI tool.

jakeefr:
Fixes #1324. Removed deprecated shell option usage that was triggering DEP0190 DeprecationWarning on Node.js 22+. I maintain PRISM (https://github.com/jakeefr/prism), a post-session diagnostics tool for Claude Code. Fixed this while testing router compatibility.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -11,7 +11,7 @@ import { runModelSelector } from "./utils/modelSelector";
 import { activateCommand } from "./utils/activateCommand";
 import { readConfigFile } from "./utils";
 import { version } from "../package.json";
-import { spawn, exec } from "child_process";
+import { spawn } from "child_process";
 import {getPresetDir, loadConfigFromManifest, PID_FILE, readPresetFile, REFERENCE_COUNT_FILE} from "@CCR/shared";
 import fs, { existsSync, readFileSync } from "fs";
 import { join } from "path";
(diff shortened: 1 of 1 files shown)


robberphex:
@musistudio same problem, please have a look. thanks.

Issue #1378, opened 2026-05-05
DeepSeek V4 Pro + thinking mode + tool calls: 400 error from API, no working workaround


GsHeri:
## Summary

When using `deepseek-v4-pro` (or `deepseek-v4-flash`) in thinking mode together with tool calls, every multi-turn request after the first tool call returns:

```
The `reasoning_content` in the thinking mode must be passed back to the API.
```

This is documented behavior on DeepSeek's side: in thinking mode, once a tool call has happened, every subsequent assistant message must carry the `reasoning_content` field on the OpenAI endpoint, or a `thinking` content block on the Anthropic endpoint. Claude Code strips both. The router currently does not re-inject them, which makes V4 unusable with Claude Code as soon as it touches tools (i.e. effectively always).

DeepSeek references:
- https://api-docs.deepseek.com/guides/thinking_mode (see "Tool Calls" section)
- https://api-docs.deepseek.com/guides/anthropic_api

## Versions

- `@musistudio/claude-code-router` `2.0.0`
- Node `v22.15.0`
- macOS 24.6.0
- DeepSeek API endpoints used:
  - OpenAI: `https://api.deepseek.com/chat/completions`
  - Anthropic: `https://api.deepseek.com/anthropic/v1/messages`

## Reproduction

Minimal `config.json`:

```json
{
  "Providers": [
    {
      "name": "d",
      "api_base_url": "https://api.deepseek.com/chat/completions",
      "api_key": "...",
      "models": ["deepseek-v4-pro"],
      "transformer": { "use": ["deepseek"] }
    }
  ],
  "Router": { "default": "d,deepseek-v4-pro" }
}
```

Multi-turn tool-use request:

```bash
curl -s -X POST http://127.0.0.1:19456/v1/messages \
  -H "Content-Type: application/json" \
  -H "x-api-key: x" \
  -H "anthropic-version: 2023-06-01" \
  -d '{
    "model": "d,deepseek-v4-pro",
    "max_tokens": 100,
    "tools": [{"name":"get_time","description":"x","input_schema":{"type":"object","properties":{}}}],
    "messages": [
      {"role":"user","content":"x"},
      {"role":"assistant","content":[{"type":"tool_use","id":"toolu_1","name":"get_time","input":{}}]},
      {"role":"user","content":[{"type":"tool_result","tool_use_id":"toolu_1","content":"12:34"}]}
    ]
  }'
```

Result:

```
[...]

PR #1381, opened 2026-05-06
fix (deepseek): proper reasoning_content passing for deepseek transformer

onokonem:
fix for https://github.com/musistudio/claude-code-router/issues/1378, works for me

please review carefully as I'm not TS programmer, just a CCR user


onokonem:
looks like this one is better than mine: https://github.com/musistudio/claude-code-router/pull/1376

PR #1430, opened 2026-06-09
Fix Dockerfile to build UI in Docker (works without out-of-context pre-build)

terafin:
Hi — thank you for maintaining claude-code-router.

## Summary

`packages/server/Dockerfile` does `COPY packages/ui/dist/. ./packages/server/dist/`, which assumes pre-built UI artifacts. That works in your release CI (which presumably runs `pnpm --filter @CCR/ui build` before `docker build`), but fails for anyone building from source on a clean clone:

```
ERROR: failed to compute cache key: failed to calculate checksum of ref ...: "/packages/ui/dist": not found
```

## What this PR does

Adds `pnpm build` for the `ui` package inside the existing `builder` stage, then `COPY --from=builder /app/packages/ui/dist/.` — same pattern already used for `core`/`shared`/`server`.

No new stages, no new tooling, no restructuring. ~7 lines of diff.

## Test plan

- `docker buildx build -f packages/server/Dockerfile --platform linux/amd64,linux/arm64 .` succeeds from a clean clone
- Resulting image starts and serves `/health` (existing healthcheck passes)
- Runtime image contents unchanged — same dist/ layout

## Backwards compatibility

Zero behavior change for your release flow. If release CI still pre-builds the UI, the Dockerfile just rebuilds it again inside Docker (adds a few seconds). Output image is byte-equivalent.

## Why this matters

Currently anyone wanting to build from a fork (local patches, downstream rebuilds, GHCR mirrors) can't `docker build` cleanly. This makes the Dockerfile self-contained — standard expectation for `docker build .` to work on a fresh clone.

### DIFF
--- a/packages/server/Dockerfile
+++ b/packages/server/Dockerfile
@@ -15,6 +15,7 @@ COPY scripts ./scripts
 COPY packages/shared/package.json ./packages/shared/
 COPY packages/core/package.json ./packages/core/
 COPY packages/server/package.json ./packages/server/
+COPY packages/ui/package.json ./packages/ui/
 
 # 安装所有依赖（包括开发依赖）并清理
 RUN pnpm install --frozen-lockfile && \
@@ -24,12 +25,16 @@ RUN pnpm install --frozen-lockfile && \
 COPY packages/shared ./packages/shared
 COPY packages/core ./packages/core
 COPY packages/server ./packages/server
+COPY packages/ui ./packages/ui
 
 # 构建所有包
 WORKDIR /app/packages/core
 RUN pnpm build
 
 WORKDIR /app/packages/shared
+RUN pnpm build
+
+WORKDIR /app/packages/ui
 RUN pnpm build
 
 WORKDIR /app/packages/server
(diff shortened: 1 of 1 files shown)


terafin:
Superseded by #1432, which extends this Dockerfile rebuild with the Node 22 + pnpm@9 fixes needed for `node:sqlite` + esbuild postinstall. Closing in favor of that PR.

terafin:
Reopening — closing this earlier was a mistake. PR #1432 (Node 20→22, pnpm@9, allow-build esbuild) has since landed but did NOT address the UI-dist COPY issue, which is the actual root cause of the build failure. They are complementary, not redundant.

Confirmed today (2026-06-10): upstream main still does `COPY packages/ui/dist/. ./packages/server/dist/` from the host build context, which fails on any clean clone with:

```
ERROR: failed to compute cache key: failed to calculate checksum of ref ...: "/packages/ui/dist": not found
```

This PR's diff (build the UI inside the builder stage, copy from there) is exactly the right fix and applies cleanly to current main.

terafin:
Withdrawing — the project moved to electron-only releases and removed packages/server/Dockerfile, so a Dockerfile change here no longer aligns with upstream's direction. We maintain our own downstream container build for internal use, which doesn't need to live upstream. Thanks for the project!

PR #1432, opened 2026-06-09
fix(docker): bump Node base 20 → 22-alpine (node:sqlite transitive breaks build)

terafin:
## Problem

`docker build -f packages/server/Dockerfile .` fails on current main with:

```
Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:sqlite
ERROR: process "/bin/sh -c pnpm install --frozen-lockfile && pnpm store prune" did not complete successfully: exit code: 1
```

`node:sqlite` is a built-in module added experimentally in Node 22.5 and stabilized later. A transitive dep in the current pnpm-lock graph requires it, so the postinstall step bombs under the `node:20-alpine` base.

## Fix

Bump both `builder` and `production` stages from `node:20-alpine` to `node:22-alpine`. Minimal, no behavioral change to runtime apart from a newer Node.

## Notes

`package.json` still declares `"engines": { "node": ">=20.0.0" }` — that's now lying about the real requirement. A follow-up could tighten it to `>=22.0.0`, but that's a separate concern and not blocking here.

## Repro

```
git checkout main
docker build -f packages/server/Dockerfile .
# → ERR_UNKNOWN_BUILTIN_MODULE: node:sqlite
```

After this PR:

```
docker build -f packages/server/Dockerfile .
# → succeeds
```

### DIFF
--- a/packages/server/Dockerfile
+++ b/packages/server/Dockerfile
@@ -1,12 +1,12 @@
 # ===========================
 # 构建阶段
 # ===========================
-FROM node:20-alpine AS builder
+FROM node:22-alpine AS builder
 
 WORKDIR /app
 
 # 安装 pnpm
-RUN npm install -g pnpm@latest && \
+RUN npm install -g pnpm@9 && \
     rm -rf /root/.npm
 
 # 复制工作区配置文件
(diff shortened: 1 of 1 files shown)


terafin:
Withdrawing — the project moved to electron-only releases and removed packages/server/Dockerfile, so a Dockerfile change here no longer aligns with upstream's direction. We maintain our own downstream container build for internal use, which doesn't need to live upstream. Thanks for the project!

PR #1446, opened 2026-06-16
fix(anthropic): re-merge inline role:system messages into last user message

ichelema:
Descrizione:

  ## Problem

  The `forEach` in `transformRequestOut` only handles `role: "user"` and
  `role: "assistant"` messages. Inline `role: "system"` messages are silently
  dropped during the Anthropic → unified conversion.

  These inline system messages are generated by Claude Code's `UserPromptSubmit`
  hooks via `additionalContext` — they carry per-prompt context such as tool
  hints and session state. They are distinct from the top-level `request.system`
  field (already handled correctly above the loop).

  Since non-Anthropic providers do not recognize inline system messages, this
  context never reaches the model when routing through any non-Anthropic provider
  (DeepSeek, OpenAI, Gemini, etc.).

  ## Fix

  Add an `else if (msg.role === "system")` branch in the `forEach` that extracts
  the text from inline system messages and appends it to the last user message,
  wrapped in an `<injected-system-context>` tag to keep it semantically distinct
  from the user's own text.

  Handles all three content formats:
  - `string` → append directly
  - `Array` of content blocks → push a new `{ type: "text" }` block
  - fallback → assign directly

  If no user message exists yet in the accumulated `messages` array, a synthetic
  user message is created so the context is never lost.

  ## Files changed

  - `packages/core/src/transformer/anthropic.transformer.ts`

### DIFF
--- a/packages/core/src/transformer/anthropic.transformer.ts
+++ b/packages/core/src/transformer/anthropic.transformer.ts
@@ -175,12 +175,13 @@ export class AnthropicTransformer implements Transformer {
           return;
         }
       } else if (msg.role === "system") {
-        // PATCH Trinity: rifondi i role:system INLINE nell'ultimo messaggio user
-        // invece di scartarli. Il forEach originale gestiva solo user/assistant, quindi
-        // i system inline (additionalContext degli hook UserPromptSubmit: skill, Hindsight)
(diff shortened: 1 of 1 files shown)


muzhi1991:
同样的问题，role system 里面的 skill 信息都丢失了，导致 skill 都识别不了。。

PR #1472, opened 2026-07-01
fix: remember main window bounds instead of forcing full-workarea size

NubeBuster:
## Summary

- The main window's initial bounds were recomputed to fill the entire display work area on *every* launch once onboarding had finished, not just the first time.
- On desktop environments that don't correctly report reserved panel/sidebar/dock space in the work area, this pins the window at full display size underneath that panel on every single launch, with no way for the user to fix it (any manual resize was discarded on the next launch).
- Now the app persists the actual window bounds (position + size) on resize/move/close and restores them on next launch. The old maximize-to-workarea behavior still runs, but only once — the first time a window is created after onboarding finishes, before any bounds have been saved.
- Stored bounds are validated against the current displays' work areas on load, so a stale position from a since-removed/reconfigured monitor can't strand the window off-screen.

## Test plan

- [x] `npm run typecheck` passes
- [ ] Manual verification on a machine with a desktop panel/sidebar that isn't reported in the Electron work area

### DIFF
--- a/src/main/windows.ts
+++ b/src/main/windows.ts
@@ -1,7 +1,8 @@
-import { app, BrowserWindow, screen } from "electron";
+import { app, BrowserWindow, screen, type Rectangle } from "electron";
 import { existsSync } from "node:fs";
 import path from "node:path";
 import { pathToFileURL } from "node:url";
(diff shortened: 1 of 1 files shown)


NubeBuster:
not possible in wayland. wtf. #the-year-of-linux

PR #1477, opened 2026-07-01
feat(session): forward x-session-id header from CCRCODE_SESSION_ID env

NubeBuster:
Inject x-session-id on outbound provider requests when CCRCODE_SESSION_ID is set — gateway path (service.ts line ~603, alongside x-client-request-id) and proxy path (createForwardHeaders, covers HTTP+HTTPS). No-op when env var absent so non-ccrcode launches are unaffected.


NubeBuster:
Allows for OpenRouter to group requests as one session

NubeBuster:
Uhhh.... triaging if this is even not already implemented...

Issue #1496, opened 2026-07-08
Local-agent OAuth provider plugin auth override never applies


emthy0:
# Local-agent OAuth provider plugin auth override never applies

Even with fresh install and setup.
<img width="1405" height="692" alt="Image" src="https://github.com/user-attachments/assets/cb2b1a6d-ef51-4387-8d4f-d872413c6d43" />

Exported config (this autogenerated, not modified): 
```json
...
  "providerPlugins": [
    {
      "auth": {
        "headers": {
          "authorization": "Bearer sk-ant-oat01-REDACT",
          "anthropic-beta": "oauth-2025-04-20"
        },
        "removeHeaders": ["x-api-key"],
        "strict": true
      },
      "key": "ccr-local-agent-claude-code-api-claude-code-oauth",
      "providerName": "Claude Code API"
    },
    {
      "auth": {
        "headers": {
          "authorization": "Bearer sk-ant-oat01-REDACT",
          "anthropic-beta": "oauth-2025-04-20"
        },
        "removeHeaders": ["x-api-key"],
        "strict": true
      },
      "key": "ccr-local-agent-claude-code-api-claude-code-oauth-internal",
      "providerName": "Claude Code API::anthropic_messages"
    },
...
]
...
```

## Summary

Providers imported from a local CLI login (Claude Code OAuth, Codex OAuth) send
requests upstream with the wrong auth header. Instead of
`Authorization: Bearer <oauth-token>` (plus `anthropic-beta: oauth-2025-04-20`),
the request goes out with `x-api-key: ccr-local-agent-login` (the internal
placeholder credential) or, after a partial fix, `x-api-key: <real oauth token>`
— still the wrong header, since Anthropic's OAuth flow requires `Authorization`,
not `x-api-key`.

## Root cause

Importing a local-agent provider (`packages/core/src/agents/local-providers/claude-code.ts`,
`codex.ts`) creates two things:

1. A `GatewayProviderConfig` entry in `config.Providers`, with `api_key` set to
   the sentinel placeholder `ccr-local-agent-login`
   (`packages/core/src/agents/local-providers/shared.ts:26`).
2. One or more `providerPlugins` entries (`bearerAuthPlugin()` /
   `apiKeyAuthPlugin()` in `shared.ts:78-114`) carrying the real captured OAuth
   token in `auth.headers.authorization`, plus `removeHeaders: ["x-api-key"]`.

The plugin is supposed to be matched to its provider by the internal gateway
[...]

PR #1497, opened 2026-07-08
fix: Local-agent OAuth provider plugin auth override never applies

emthy0:
# Local-agent OAuth provider plugin auth override never applies

Even with fresh install and setup.
Image

Exported config (this autogenerated, not modified): 
```json
...
  "providerPlugins": [
    {
      "auth": {
        "headers": {
          "authorization": "Bearer sk-ant-oat01-REDACT",
          "anthropic-beta": "oauth-2025-04-20"
        },
        "removeHeaders": ["x-api-key"],
        "strict": true
      },
      "key": "ccr-local-agent-claude-code-api-claude-code-oauth",
      "providerName": "Claude Code API"
    },
    {
      "auth": {
        "headers": {
          "authorization": "Bearer sk-ant-oat01-REDACT",
          "anthropic-beta": "oauth-2025-04-20"
        },
        "removeHeaders": ["x-api-key"],
        "strict": true
      },
      "key": "ccr-local-agent-claude-code-api-claude-code-oauth-internal",
      "providerName": "Claude Code API::anthropic_messages"
    },
...
]
...
```

## Summary

Providers imported from a local CLI login (Claude Code OAuth, Codex OAuth) send
requests upstream with the wrong auth header. Instead of
`Authorization: Bearer <oauth-token>` (plus `anthropic-beta: oauth-2025-04-20`),
the request goes out with `x-api-key: ccr-local-agent-login` (the internal
placeholder credential) or, after a partial fix, `x-api-key: <real oauth token>`
— still the wrong header, since Anthropic's OAuth flow requires `Authorization`,
not `x-api-key`.

## Root cause

Importing a local-agent provider (`packages/core/src/agents/local-providers/claude-code.ts`,
`codex.ts`) creates two things:

1. A `GatewayProviderConfig` entry in `config.Providers`, with `api_key` set to
   the sentinel placeholder `ccr-local-agent-login`
   (`packages/core/src/agents/local-providers/shared.ts:26`).
2. One or more `providerPlugins` entries (`bearerAuthPlugin()` /
   `apiKeyAuthPlugin()` in `shared.ts:78-114`) carrying the real captured OAuth
   token in `auth.headers.authorization`, plus `removeHeaders: ["x-api-key"]`.

The plugin is supposed to be matched to its provider by the internal gateway
process (`@the-next-ai/ai-gateway`, config written by `writeCoreGatewayConfig()`
in `packages/core/src/gateway/service.ts:1151`) via an **exact string match**
between the plugin's `providerName` field and the provider's *runtime* name.

The runtime name is computed by `providerRuntimeId()`
(`gateway/service.ts:6671`):

```ts
function providerRuntimeId(provider) {
  const explicit = sanitizeProviderHeaderId(provider.id);
  if (explicit) return explicit;
  // ...falls back to `provider-<slugified-name>-<sha256-hash-of-name+baseUrl>`
}
```

`provider.id` was **never set** during import
(`packages/ui/src/pages/home/App.tsx`, provider-save handler around line 1434
— no `id` field on the constructed `GatewayProviderConfig`). So the backend
always fell into the hash branch, producing an opaque name like
`provider-claude-code-api-884b99c439::anthropic_messages`.

Meanwhile, the plugin's `providerName` was set (in
`materializeProviderPluginTemplates()`, `App.tsx:60-75`) to the **human-readable
label**: `"Claude Code API::anthropic_messages"`.

These two strings never match. The gateway's plugin-resolution step
(`resolve(n, t)` in the vendor bundle) silently no-ops — no error, no log —
and the provider's raw `api_key` (the sentinel, or later the swapped-in real
token) goes out using the protocol's default header convention. For Anthropic
(`type: "anthropic"`), that default is always `x-api-key`, never `Authorization`,
regardless of `extraHeaders`.


emthy0:
#1496

PR #1499, opened 2026-07-08
fix: Update default sonnet model as old one is no longer existed (404)

emthy0:
Anthroipic server no longer have old claude-sonnet-4-20250514
Refer to [Claude Legacy Model](https://platform.claude.com/docs/en/about-claude/models/overview#:~:text=Claude%20Sonnet%204.6-,Claude%20Sonnet%204.5,-Claude%20Opus%204.5)

image


emthy0:
Update 
Legacy model may cause incompatible with latest claude without proper handling, Update to latest model instead
image
image

PR #1503, opened 2026-07-09
fix: Kimi Global Provider's Endpoint Appended with Anthropic

bytrangle:
When adding provider Kimi Global, the endpoint is automatically appended with `/anthropic` endpoint even though it uses OpenAI API format. The final url is wrong and therefore the Kimi Global provider can't be added.

https://github.com/user-attachments/assets/5340b2eb-a4f2-40bc-a353-a555e39be562

**Causes**
In `providerProbeCandidates`, it replaces the declared protocol with every protocol known to the app.
`shouldProbeAnthropicPrefixFallack` at `providers/url.ts` is overly broad. It returns true for any URL that didn't already have "anthropic" in the path. This means the url becomes `https://api.moonshot.ai/anthropic`. This endpoint responds with 401 (auth required), so the probe mis-detects it as a valid Anthropic endpoint.

**Solution**
When a preset is selected, use the endpoint's own declared protocols instead of probing every protocol against every endpoint.
Fallback candidates are only generated when the hostname of path already contains "anthropic".

### DIFF
--- a/packages/core/src/providers/url.ts
+++ b/packages/core/src/providers/url.ts
@@ -137,12 +137,18 @@ function shouldProbeOpenAiV1Fallback(value: string): boolean {
 }
 
 function shouldProbeAnthropicPrefixFallback(value: string): boolean {
+  // Only add /anthropic fallback when the URL already shows anthropic affinity:
+  // - hostname contains "anthropic" (e.g. api.anthropic.com)
+  // - path contains "anthropic" (e.g. api.z.ai/api/anthropic)
+  // This prevents false positives where a non-anthropic endpoint
+  // coincidentally responds to /anthropic/* with a 401/403.
   const url = new URL(value);
   const segments = url.pathname
     .split("/")
     .map((segment) => segment.trim().toLowerCase())
     .filter(Boolean);
-  return !segments.includes("anthropic");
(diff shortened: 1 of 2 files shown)


musistudio:
I think this should be a configuration error after the probe protocol. The current project already supports configuration like:
```
"capabilities": [
  { "type": "openai_chat_completions", "baseUrl": "https://api.moonshot.ai/v1" },
  { "type": "anthropic_messages", "baseUrl": "https://api.moonshot.ai/anthropic" }
]
```
So you just need to configure it following this approach.I will fix it in the next release.

bytrangle:
Unless I'm mistaken, the UI flow is:
1. Select Kimi Global preset -> form shows base url https://api.moonshot.ai/v1 (correct)
2. Enter API key -> probe runs -> `normalizeBaseUrl` returns https://api.moonshot.ai/anthropic (incorrect)
3. `applyProviderProbeResult` overwrites the form's base url with the value in step 2

I can't edit the base url because the form field is uneditable.
Do I miss anything?

musistudio:
What I mean is that this happened because the detection logic mistakenly overwrote the base URL with the Anthropic protocol. In the next version, all protocols will still be probed, but the Anthropic protocol will have its own base URL, separate from OpenAI.

PR #1524, opened 2026-07-11
fix: strip Provider/ prefix from model name before sending to ai-gateway

810senpai114514:
## Summary

When body.model is `NebulaCoder/nebulacoder-cot-v8.0` (Provider/Model format), the Provider/ prefix is not stripped before sending to ai-gateway, causing a 400 error.

## Bug evidence from request_logs.sqlite

Request body:
```json
{"model":"NebulaCoder/nebulacoder-cot-v8.0", ...}
```

Response:
```json
{"error":{"message":"All target providers failed.",
  "attempts":[{"provider":"openai",
    "message":"Model \"NebulaCoder/nebulacoder-cot-v8.0\" is not configured
               for target provider openai.
               Allowed models: nebulacoder-v8.0, nebulacoder-cot-v8.0."}]}}
```

## Root cause

`normalizeConfiguredProviderModelBody` correctly strips the prefix, but `prepareUpstreamCredentialAttempt`'s `!target` branch preferred the original `input.attempt.body` (with `NebulaCoder/` prefix) over `normalizedBody.body` when the body was parseable.

## Fix

Always use `normalizedBody.body` (which has the provider prefix stripped) when available. Remove the now-unused `bodyHasConfiguredProviderModelSelector` function.

### DIFF
--- a/packages/core/src/gateway/service.ts
+++ b/packages/core/src/gateway/service.ts
@@ -5775,9 +5775,7 @@ function prepareUpstreamCredentialAttempt(input: {
     target
   });
   if (!target) {
-    const body = bodyHasConfiguredProviderModelSelector(input.attempt.body, input.config)
-      ? input.attempt.body
-      : normalizedBody?.body ?? input.attempt.body;
+    const body = normalizedBody?.body ?? input.attempt.body;
     return {
       ...input.attempt,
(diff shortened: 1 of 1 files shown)


810senpai114514:
Closing to recreate with better description

PR #1548, opened 2026-07-15
fix(gateway): align provider plugin runtime identity

camjac251:
## Summary

- normalize provider-plugin targets to the runtime or capability identity emitted by gateway compilation
- use the same runtime identity for credential-free fallback headers
- preserve unmatched and unscoped plugin entries unchanged

## Why

Provider compilation replaces display and legacy capability names with stable runtime identifiers. Plugin targets and fallback headers could still carry the precompiled name, so authentication rules were registered against an identity the gateway never selected. This change keeps those boundaries aligned.

## Test plan

- [x] core unit suite - 259 passed, 2 skipped
- [x] `TZ=UTC npm run test:main` - core: 417 passed, 5 skipped; electron: 2 passed
- [x] `npm run typecheck`
- [x] `npm run build:assets`
- [x] `git diff --check`

### DIFF
--- a/packages/core/src/gateway/core-runtime/config-compiler.ts
+++ b/packages/core/src/gateway/core-runtime/config-compiler.ts
@@ -40,20 +40,16 @@ export async function compileCoreGatewayConfig(
       )
     : [];
   const pluginBillingConfig = isRecord(pluginCoreGatewayConfig.billing) ? pluginCoreGatewayConfig.billing : {};
-  const configuredProviderPlugins = normalizeClaudeCodeOauthProviderPlugins(
-    normalizeCoreProviderPluginNames(
-      [
-        ...(config.providerPlugins ?? []).filter(providerPluginEnabled),
-        ...pluginService.getCoreProviderPlugins().filter(providerPluginEnabled)
-      ],
-      config.Providers
-    )
-  );
-  const providerPlugins = await withKimiOauthRuntimeDefaults(
+  const configuredProviderPlugins = normalizeClaudeCodeOauthProviderPlugins([
+    ...(config.providerPlugins ?? []).filter(providerPluginEnabled),
+    ...pluginService.getCoreProviderPlugins().filter(providerPluginEnabled)
+  ]);
+  const providerPluginsWithRuntimeDefaults = await withKimiOauthRuntimeDefaults(
     await withGrokOauthRuntimeDefaults(withCodexOauthRuntimeDefaults(configuredProviderPlugins))
   );
+  const codexOauthProviderNames = codexOauthLocalProviderNames(providerPluginsWithRuntimeDefaults);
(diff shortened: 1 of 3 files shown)


packages/core/src/gateway/core-runtime/config-compiler.ts
```diff
@@ -29,10 +29,15 @@ export async function compileCoreGatewayConfig(
 ): Promise<Record<string, unknown>> {
   const pluginCoreGatewayConfig = pluginService.getCoreGatewayConfig();
   const pluginBillingConfig = isRecord(pluginCoreGatewayConfig.billing) ? pluginCoreGatewayConfig.billing : {};
-  const configuredProviderPlugins = normalizeClaudeCodeOauthProviderPlugins([
-    ...(config.providerPlugins ?? []).filter(providerPluginEnabled),
-    ...pluginService.getCoreProviderPlugins().filter(providerPluginEnabled)
-  ]);
+  const configuredProviderPlugins = normalizeClaudeCodeOauthProviderPlugins(
```
musistudio: This normalization now runs before `codexOauthLocalProviderNames()` builds the set used by `withCodexOauthProviderBaseUrl()`. For local Codex OAuth plugins, `providerName: "Codex API"` can be rewritten to the runtime id such as `codex-api`, but `withCodexOauthProviderBaseUrl()` later checks only `codexOauthProviderNames.has(provider.name)`. A provider whose display name is still `"Codex API"` will no longer match, so the Codex OAuth provider keeps its original `api_base_url` instead of being rewritten to `codexDefaultBaseUrl`. Please either preserve/match both display and runtime names in that Codex OAuth path, or move this normalization after the Codex OAuth base-url detection, and add a regression test that asserts a Codex OAuth plugin normalized to a runtime id still produces a provider with `baseurl === codexDefaultBaseUrl`.

musistudio (changes requested):
Thanks for tightening the provider runtime identity boundary. I found one regression in the Codex OAuth path: normalizing the plugin providerName before collecting Codex OAuth provider names means the later provider base-url rewrite can miss the provider, because it still matches only against the display name. That leaves local Codex OAuth providers pointed at the old base URL while the OAuth plugin is registered under the runtime id. I think this needs a small fix and a regression test before merging.

packages/core/src/gateway/core-runtime/config-compiler.ts
camjac251: Addressed in 7dc7b82. The compiler now captures local OAuth provider display names after runtime defaults but before plugin identity normalization, so the provider rewrite still matches while the emitted plugin retains its runtime ID. Added a regression asserting both the normalized plugin name and managed default `baseurl`. Verified the focused file (3/3), complete main suite (298 passed, 5 skipped), typecheck, and production asset build.

PR #1550, opened 2026-07-15
fix: translate Claude effort for OpenAI Responses

jesieleo:
## Summary

- Translate Claude `output_config.effort` into OpenAI Responses `reasoning.effort`.
- Use target-model metadata to preserve or safely downgrade unsupported reasoning levels.
- Preserve an existing native `reasoning` configuration.
- Leave `output_config` unchanged for chat-completions targets so compatibility plugins can still consume it.

## Why

Claude Code can send reasoning effort as:

```json
{
  output_config: {
    effort: xhigh
  }
}
```

When that request was routed to an OpenAI Responses provider, the field could be forwarded unchanged and the upstream returned HTTP 400:

```text
Unsupported parameter: output_config
```

This change normalizes the parameter at CCR's upstream boundary before the protocol adapter builds the Responses request.

## Effort handling

- Supported levels are preserved: `low`, `medium`, `high`, `xhigh`, and `max`.
- If the target model does not support the requested level, CCR selects the closest supported level without exceeding it when possible.
- Without model metadata, standard Responses levels are used, so `max` falls back to `xhigh`.
- Unknown effort values are removed and the target model uses its default.
- An explicitly supplied native `reasoning.effort` takes precedence.

## Validation

- `npm.cmd run typecheck`
- `npm.cmd run build:assets`
- `node --test dist/tests/main/router-builtins.test.js` — 46 tests passed
- Manually verified the original Claude Code → OpenAI Responses request no longer fails with `Unsupported parameter: output_config`.

### DIFF
--- a/packages/core/src/gateway/upstream/executor.ts
+++ b/packages/core/src/gateway/upstream/executor.ts
@@ -23,6 +23,12 @@ import { claudeCodeOauthBetaHeader, claudeCodeOauthRequiredBeta, UpstreamRequest
 import type { ApiKeyLimitUsage, ProviderCredentialRoutingTarget, UpstreamAttempt, UpstreamFailedAttempt, UpstreamFetchResult } from "@ccr/core/gateway/internal/shared";
 
 const providerCredentialSpilloverThreshold = 0.8;
+const openAiResponsesReasoningEffortOrder = ["none", "minimal", "low", "medium", "high", "xhigh", "max"] as const;
+const defaultOpenAiResponsesReasoningEfforts = new Map(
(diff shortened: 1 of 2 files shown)


musistudio:
I think this is most likely an issue with the upstream dependency. Would you be able to submit a PR to the upstream project at https://github.com/The-NeXT-AI/ai-gateway?

jesieleo:
> 我认为这很可能是上游依赖项的问题。您能否向上游项目提交一个 PR，地址是https://github.com/The-NeXT-AI/ai-gateway？

收到，我把这部分改到 ai-gateway 里，再去那边提一个 PR。

PR #1552, opened 2026-07-16
fix(gateway): honor provider context limits

camjac251:
## Summary

- derive effective input windows from provider metadata and its configured percentage
- use maximum provider capacity for optional one-million-context capability while keeping normal token budgets on the effective window
- keep published routes, capabilities, inference hints, subagent routing, and display labels consistent
- deduplicate routes after applying effective limits
- remove an unreachable alternate discovery builder

## Why

The static catalog can disagree with the provider that actually serves a model. A bounded provider could be advertised with a larger normal window, while an uncatalogued provider with a larger configured capacity could lose an optional route.

Effective input limits and physical maximum capacity are separate. A provider can use a bounded normal window while still supporting an explicitly selected one-million-context variant. Provider metadata is authoritative when present, and catalog values remain the fallback when metadata is absent.

## Test plan

- [x] focused gateway model-discovery coverage, including bounded normal windows and larger optional capacity
- [x] core unit suite: 265 passed, 2 skipped, 0 failed
- [x] typecheck
- [x] production asset build
- [x] whitespace validation
- [x] secret scan
- [ ] full main suite: 416 passed and 5 skipped; one unrelated observability bounded-heap assertion failed under Node 24.18.0 because its generated body did not exceed the reported heap limit

### DIFF
--- a/packages/core/src/agents/claude-app/gateway-routes.ts
+++ b/packages/core/src/agents/claude-app/gateway-routes.ts
@@ -1,6 +1,7 @@
 import type { AppConfig } from "@ccr/core/contracts/app";
 import { availableGatewayModelIds, normalizeProfileScopeValue } from "@ccr/core/contracts/app";
 import { modelRegistryForConfig } from "@ccr/core/routing/model-registry";
+import { resolveUsageModelAttribution } from "@ccr/core/usage/model-attribution";
 
 export const CLAUDE_APP_ONE_MILLION_CONTEXT_SUFFIX = "[1m]";
 const CLAUDE_APP_ENCODED_ROUTE_PREFIX = "anthropic/claude-ccr-h";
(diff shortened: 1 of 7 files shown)


packages/core/src/agents/claude-app/gateway-routes.ts
```diff
@@ -122,6 +127,40 @@ export function stripClaudeAppGatewayOneMillionContextSuffix(id: string): string
   return id.trim().replace(/\[1m\]$/i, "").trim();
 }
 
+export function effectiveProviderModelContextWindow(
+  config: Pick<AppConfig, "Providers" | "virtualModelProfiles">,
+  model: string
+): number | undefined {
+  const resolved = modelRegistryForConfig(config).resolveProviderModel(
```
musistudio: **[P2] Resolve virtual-model routes before reading provider metadata**

resolveProviderModel() returns only provider-kind resolutions, while visible prefix, suffix, and exact-alias virtual models are registered as gateway models. As a result, this helper returns undefined for those published routes even when their underlying provider model has context metadata.

I reproduced this with a 1,000,000-token Gateway/custom-frontier model and a visible tools- virtual prefix. The resulting Gateway/tools-custom-frontier route reported max_input_tokens: 0 and oneMillionContext: false. Please resolve the virtual selector to its physical provider model before reading metadata, and add coverage for prefix/suffix and fixed-alias profiles.

camjac251: Thanks for catching the virtual-route gap. Addressed in `a3889d7`. Context lookup now keeps explicit provider routes authoritative, then resolves prefix, suffix, and fixed-alias routes to their physical provider model before reading metadata. Regressions cover all three virtual selectors, cyclic fallback, and a physical-route/exact-alias collision. Verified with typecheck, 264 passing core unit tests with 2 skips, and the production asset build.

Issue #575, opened 2025-08-14
Provider + Model picker?


akamalov:
Greetings,

Wondering is it  possible to select a provider + model while running `ccr code` ?  Let's say I want to select Kimi-k2, which I have configured in my `config.json`:
```
{
            "name": "kimi-k2",
            "api_base_url": "https://openrouter.ai/api/v1",
            "api_key": "sk-or-v1-e08XXXXXXXXXXXXXXXXXXX",
            "models": [
                "mooonshotai/kimi-k2"
            ]
       }
...

```
Will it be possible to select it from within  Claude Code, or do I need to relaunch it pointing to Provider/Model. If the latter, then how ?

Thanks again!

echoyolo6:
<img width="868" height="388" alt="Image" src="https://github.com/user-attachments/assets/f6e33305-e210-4c79-9684-a6dce07fb6a4" />

Issue #504, opened 2025-08-09
/model does not show configured models


V4G4X:
From the demos I have seen on Youtube. ClaudeCode's model picker shows the model list from our custom providers, right?

I have configured that the above is working, I see Activity on OpenRouter when I use `ccr code`.
But the `/model` picker does not show these options.
It shows the default `sonnet` and `opus` models.

<img width="777" height="332" alt="Image" src="https://github.com/user-attachments/assets/96635423-fbc4-4be8-9913-a7b393b9a6db" />

Config: 
```json
{
  "LOG": false,
  "CLAUDE_PATH": "",
  "HOST": "127.0.0.1",
  "PORT": 3456,
  "APIKEY": "",
  "API_TIMEOUT_MS": "600000",
  "PROXY_URL": "",
  "transformers": [],
  "Providers": [
    {
      "name": "openrouter",
      "api_base_url": "https://openrouter.ai/api/v1/chat/completions",
      "api_key": "my-key",
      "models": [
        "openai/gpt-5",
        "openai/o3",
        "google/gemini-2.5-flash-lite",
        "google/gemini-2.5-pro",
        "anthropic/claude-sonnet-4"
      ],
      "transformer": {
        "use": [
          "openrouter"
        ]
      }
    }
  ],
  "Router": {
    "default": "openrouter,openai/gpt-5",
    "background": "openrouter,google/gemini-2.5-flash-lite",
    "think": "openrouter,openai/o3",
    "webSearch": "openrouter,google/gemini-2.5-pro"
  }
}
```

Is this a known issue that broke with a recent ClaudeCode update?

JJsilvera1:
You have to type it in the first time

/Models openrouter,google/gemini-2.5-flash-lite

You have to do this every time you switch a model though

guyjames:
> You have to type it in the first time
> 
> /Models openrouter,google/gemini-2.5-flash-lite
> 
> You have to do this every time you switch a model though

It's actually `/model openrouter,google/gemini-2.5-flash-lite`

V4G4X:
This helps. At least now I'm able to switch on the fly.

But do you think it would be possible to get it to show all the models that we configured in `Providers`?

JJsilvera1:
I wish. But no I don't think it's possible I have to change the main model in the settings for it to work. It's a two-step process

PR #1557, opened 2026-07-17
feat(profiles): constrain selectable models

camjac251:
## Summary

- add an optional `allowedModels` policy to managed CLI profiles
- compile native aliases, version prefixes, and configured provider selectors into the exact IDs published by gateway discovery
- inject the policy only for the selected profile launch without rewriting inherited or isolated user settings
- remove duplicate provider discovery rows only for an authenticated matching CLI profile while retaining explicitly selected routes
- share generated profile-key identity across launch, authentication, persistence, and discovery
- add UI controls, English and Chinese documentation, lifecycle coverage, and input-boundary tests

## Stack

This change depends on:

- #1552 for provider context metadata and maximum-capacity route support
- #1554 for inherited configuration and the launch-wrapper foundation

Until both prerequisites merge, their six commits also appear in this branch. The profile policy is commit `3822496`; the discovery refinement is commit `5ccfd40`.

## Behavior

- an empty field preserves the client's existing availability
- unknown, ambiguous, stale, or unsupported selectors fail during profile application
- configured defaults must be included in the profile policy
- ordinary `--settings` remains available, while an explicit competing `--managed-settings` option is rejected
- existing enterprise policy remains authoritative
- the normal launcher and the user's existing settings file remain unchanged
- duplicate provider rows covered by native aliases are hidden only from the discovery response
- explicitly selected provider routes remain visible, and hidden routes remain available to inference
- unrelated keys, app-surface profiles, non-CCR profiles, and globally disabled profiles keep the complete discovery response

Related to #504 and #575.

## Validation

- [x] monorepo typecheck
- [x] focused discovery, routing, and inference suite: 81 passed, 0 failed
- [x] core unit suite: 279 passed, 2 skipped, 0 failed
- [x] changed profile integration file: 34 passed, 0 failed
- [x] changed wrapper integration file: 22 passed, 3 skipped, 0 failed
- [x] UI component suite: 59 passed, 0 failed
- [x] production asset build
- [x] English and Chinese documentation lint and static build
- [x] staged secret scan
- [x] current client binary accepts the launch-only policy option with a native version prefix

A full core sweep under the default local runtime reached 473 passed and 7 skipped. One unrelated bounded-heap regression outside the changed files failed because its generated payload did not exceed the runtime's reported heap limit. Every changed discovery, routing, profile, and wrapper test passed.


musistudio:
If we support the `allowedModels`, what problems can this help us solve?

camjac251:
The main use case is per-profile model scoping. A profile can have many gateway-discovered routes, including aliases and routes intended for other profiles, while only a smaller set should be selectable for that launch.

`allowedModels` limits `/model` and explicit subagent model overrides to that curated set, rejects unknown or ambiguous selectors before launch, and verifies that configured default and small-fast model selections are included when provided.

It does not remove backend routing capability. Hidden discovery entries can still support route inference, explicitly selected provider routes remain visible, and existing enterprise policy remains authoritative. Leaving the field unset preserves the current behavior.

musistudio:
> The main use case is per-profile model scoping. A profile can have many gateway-discovered routes, including aliases and routes intended for other profiles, while only a smaller set should be selectable for that launch.
> 
> `allowedModels` limits `/model` and explicit subagent model overrides to that curated set, rejects unknown or ambiguous selectors before launch, and verifies that configured default and small-fast model selections are included when provided.
> 
> It does not remove backend routing capability. Hidden discovery entries can still support route inference, explicitly selected provider routes remain visible, and existing enterprise policy remains authoritative. Leaving the field unset preserves the current behavior.

That sounds reasonable. I'll review the implementation plan and code later.

PR #1578, opened 2026-07-22
Add Kenari provider preset

doedja:
## What

Adds a provider preset for [Kenari](https://kenari.id/), an Indonesian LLM gateway (OpenAI-compatible API with local-currency billing).

- `packages/core/src/providers/presets/kenari/index.ts`: preset with `https://kenari.id/v1` serving `anthropic_messages`, `openai_chat_completions`, and `openai_responses` on the same base URL
- Registered in `presets/index.ts`

## Notes

- Kenari's models are already present in `packages/core/models.json` via the models.dev sync (provider id `kenari`, aliases like `kenari/claude-sonnet-5`); this preset adds the matching provider entry
- API key format is `kn-` + 48 hex chars, covered by `officialApiKeyPatterns`
- Live model list with pricing and context length: `GET https://kenari.id/v1/models` (no auth required)
- Docs: https://kenari.id/docs

## Testing

- `npm run test:core`: 547 tests, 542 pass, 0 fail (5 skipped, pre-existing)
- `npm run test:architecture`: 4 pass, 0 fail

Follows the shape of the MiniMax preset PR (#1504).

### DIFF
--- a/packages/core/src/providers/presets/index.ts
+++ b/packages/core/src/providers/presets/index.ts
@@ -6,6 +6,7 @@ import { deepSeekProviderPreset } from "@ccr/core/providers/presets/deepseek/ind
 import { fennoProviderPreset } from "@ccr/core/providers/presets/fenno/index";
 import { geminiProviderPreset } from "@ccr/core/providers/presets/gemini/index";
 import { infistarAiProviderPreset } from "@ccr/core/providers/presets/infistar-ai/index";
+import { kenariProviderPreset } from "@ccr/core/providers/presets/kenari/index";
 import { kimiCodingProviderPreset } from "@ccr/core/providers/presets/kimi-coding/index";
 import { minimaxChinaProviderPreset, minimaxGlobalProviderPreset } from "@ccr/core/providers/presets/minimax/index";
 import { mistralProviderPreset } from "@ccr/core/providers/presets/mistral/index";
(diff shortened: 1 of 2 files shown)


musistudio:
Could you add an "Import to CCR" button on your site? You can refer to the documentation: https://ccrdesk.top/en/configuration/provider-deeplink/

doedja:
Both things are done.

**The import button is live**, thanks for pointing me at the deeplink docs: https://kenari.id/docs#tools

The link it produces:

```
ccr://provider?name=Kenari&base_url=https%3A%2F%2Fkenari.id%2Fv1&protocol=anthropic_messages&icon=https%3A%2F%2Fkenari.id%2Fbrand%2Fmark.svg&source=https%3A%2F%2Fkenari.id%2F
```

Two deliberate choices there. It is a plain anchor rather than the `ccr-provider-buttons.js` embed, to avoid putting a third-party script on a customer-facing page. And `models` is omitted on purpose: your bundled `packages/core/models.json` already carries Kenari through the models.dev sync, and the probe resolves the live list after the key is entered, so a hardcoded list in the URL would just go stale.

**This PR is rebased** onto current main and mergeable again. Still the same two files, 21 lines: the preset and its registry entry. I checked it against the current `ProviderPreset` shape and re-verified the three `defaultModels` against the live catalog.

On tests: `test:core` has three failing plugin tests (Claude Design plugin config and bundled plugin permissions), but they fail identically on a clean checkout of main with this branch absent, so they are not from this change.

PR #1579, opened 2026-07-22
Refresh MiniMax model metadata and endpoints

octo-patch:
Reason: Refresh MiniMax model metadata, pricing, and regional endpoints to match current official parameters.

This change:

- normalizes MiniMax-M3 and MiniMax-M2.7 context windows, input modalities, and official pricing in the generated model catalog
- maps the MiniMax regional presets to their catalog provider records
- includes both current models in the global and China presets
- corrects the Anthropic-compatible base URLs for both regions
- adds focused coverage for model metadata, pricing, models, and endpoints

Checks:

- `node --check scripts/generate-models-json.mjs`
- `npm run typecheck`
- `npm run test:unit -w @claude-code-router/core` (367 tests: 366 passed, 1 skipped)
- `git diff --check`


musistudio:
A Minimax official once said in a public setting that you should not use the 1M context window unless absolutely necessary. I'm not sure whether this is appropriate.

octo-patch:
Thanks for flagging this. The 1,000,000-token value records MiniMax-M3's advertised maximum capability in the catalog; it does not make the router select or encourage that context size by default, so the operational recommendation remains unchanged.

musistudio:
> Thanks for flagging this. The 1,000,000-token value records MiniMax-M3's advertised maximum capability in the catalog; it does not make the router select or encourage that context size by default, so the operational recommendation remains unchanged.

But why do we need to modify models.json? This is automatically generated by the script.

octo-patch:
Thanks for the review. I removed the generated packages/core/models.json change, kept the MiniMax generator and preset updates, and merged the current main branch. I ran node --check on the generator and the focused provider catalog and preset tests (15 passed).

octo-patch:
Thanks for the earlier review. I kept the generated catalog out of the diff and added explicit thinking-mode overrides so MiniMax-M3 remains adaptive/disable-capable and MiniMax-M2.7 remains always-on when the catalog is regenerated. I ran the generator assertions, the focused provider tests (15 passed), and npm run typecheck.

PR #1604, opened 2026-07-29
fix(claude-code): look up the macOS keychain item by account

### LINKED ISSUE #1567: Does this work with subscription usage or only APIs?
Getting errors after running this.

### LINKED ISSUE #1601: Claude Code login import fails silently on CC 2.1.220 (macOS): keychain service is now suffixed and holds no `claudeAiOauth`
## Summary

On macOS with Claude Code **2.1.220**, the "Import Local Agent Login → Claude Code" path cannot find credentials, and the failure is silent. Root cause is a storage-layout change in Claude Code, not a keychain permission problem.

Related but distinct from #1567 (there the import *was* detected and failed later at 401 / expired token). Here the credentials are never located at all.

## Environment

- CCR **3.0.17** (npm `@musistudio/claude-code-router`, verified installed `dist` matches source)
- macOS (Darwin 25.5.0), Apple Silicon
- Claude Code **2.1.220**, subscription (OAuth) login, freshly re-authenticated via `/login`

## What CCR looks for

`packages/core/src/agents/local-providers/claude-code.ts`

```ts
const claudeCodeKeychainService = "Claude Code-credentials";   // line 24
```

plus `claudeCredentialFiles()`:
- `~/.claude/.credentials.json`
- `~/.claude/credentials.json`
- `~/.config/claude/credentials.json`

## What Claude Code 2.1.220 actually does

**1. The keychain service name is now suffixed.** `security dump-keychain` shows:

```
"svce"<blob>="Claude Code-credentials"            <- legacy item, still present
"svce"<blob>="Claude Code-credentials-<8hex>"
"svce"<blob>="Claude Code-credentials-<8hex>"
```

CCR's hardcoded unsuffixed name only matches the legacy item. There is no glob/prefix match and no env override (`CCR_CLAUDE_*` vars target the app binary and config dir, not credentials).

**2. More importantly, none of those items contains an OAuth token.** All three parse as JSON and contain only MCP plugin OAuth:

```
top-level keys: ['mcpOAuth']
  mcpOAuth -> ['plugin:<name>|<id>', ...]
```

No `claudeAiOauth`, and no `accessToken` / `access_token` / `anthropicAccessToken` at any depth — so `findOauthTokenSet()` (`shared.ts:133`) correctly returns `undefined`.

**3. None of the three fallback files exists.** `~/.claude.json` holds only config/state (`userID`, migration flags, caches) — no credential material.

So the read succeeds (`security ... -w` exits 0, valid JSON) and parsing succeeds; there is simply no token in any location CCR knows about. Claude Code 2.1.220 appears to store its OAuth token somewhere outside all of these paths.

## Secondary observation

Docs state: *"If CCR only detects login traces but no usable access token, the import entry shows why it cannot be imported."* In this case no explanatory entry was visible in **Add Provider**. I'm less certain about this one — it may have been present and overlooked — but if the candidate is suppressed when the token is missing, surfacing the reason would have made this diagnosable in seconds instead of a long investigation. `readClaudeCodeKeychainRecord()` also discards stderr (`stdio: ["ignore","pipe","ignore"]`) and swallows all errors via `catch {}`, which hides the cause.

## Suggested direction

1. Match `Claude Code-credentials*` (prefix/glob) rather than the exact legacy name, and prefer the most recently modified item.
2. Surface a diagnostic when an item is found but yields no token set (distinct from "nothing found").

Happy to test patches — I have a reproducible environment.

matteoantoci:
Fixes #1601. Likely also fixes #1567 — see the `mcpOAuth` note below.

## Problem

On macOS with Claude Code >= 2.1, "Import Local Agent Login → Claude Code" finds no credentials and fails silently.

Claude Code writes its credential item under the current `$USER`, and leaves any pre-2.1 item (account `unknown`) in place on the **same** service name, `Claude Code-credentials`. `readClaudeCodeKeychainRecord()` runs:

```
security find-generic-password -s "Claude Code-credentials" -w      # no -a
```

With no `-a`, the Keychain returns an arbitrary item among those sharing the service — in practice the stale legacy one, which carries only `mcpOAuth`. The read exits 0, the JSON parses, `findOauthTokenSet()` correctly returns `undefined`, and `service.ts` then drops the candidate entirely via `.filter(c => c.status !== "missing")`. Hence: no token, no error, no entry in Add Provider.

`security dump-keychain | grep svce` de-duplicates the two same-named items, which makes this easy to misdiagnose as a service-name-suffix problem. It isn't — on a default install the service name is exactly the legacy one CCR already uses. **The account is the only thing missing in the common case.**

Full analysis, including the naming algorithm decompiled out of the 2.1.220 binary: https://github.com/musistudio/claude-code-router/issues/1601#issuecomment-5116055653

## Change

`packages/core/src/agents/local-providers/claude-code.ts` only (plus tests).

**Tiered candidate resolution**, lazily evaluated so the normal case stays at one `security` call:

1. the expected service name, derived the way Claude Code derives it — `` `Claude Code${oauthSuffix}-credentials${configSuffix}` ``, where `configSuffix` is `-${sha256(NFC(configDir)).slice(0, 8)}` once `CLAUDE_CONFIG_DIR` / `CLAUDE_SECURESTORAGE_CONFIG_DIR` is set — read under the current account;
2. `security dump-keychain` enumeration, newest-modified first, for items under another account or a config dir this process cannot reconstruct. Without `-d` that prints item **metadata only**: it never decrypts a password and never prompts;
3. the previous accountless read, as a last resort.

**`claudeAiOauth` wins over the recursive search.** `mcpOAuth` holds per-plugin OAuth records that also carry `accessToken`, and key order puts it first — so a recursive match can import an MCP plugin token as an Anthropic provider that then 401s on every request. Recursive search is now a fallback used only when no candidate has an explicit `claudeAiOauth`. This is why I think it also covers #1567.

**Diagnostics instead of silence.** `security` stderr is captured rather than discarded via `stdio: ["ignore","pipe","ignore"]`, and a `locked` candidate naming the reason is returned when login state exists but yields no token, so it survives the `status !== "missing"` filter. `errSecItemNotFound` (exit 44) is treated as absence, not an error, so a machine with no Claude Code login still correctly reports `missing` rather than a confusing `locked`.

**File fallback fixed.** `continue` past a credentials file that holds no token instead of returning it, and check the secure-storage config dir, which the env vars above can relocate away from `~/.claude`.

No public API changed. `readClaudeCodeOauth()` keeps its signature; `scanClaudeCodeLogin()` is added for the diagnostic path.

## Verification

- **9/9** in `local-agent-provider-claude-code.test.mjs`. 6 are new: per-account lookup (this bug), `dump-keychain` enumeration for an item under another account, `mcpOAuth` rejection in favour of `claudeAiOauth`, config-dir-suffixed service name, the tokenless diagnostic, and logged-out-stays-`missing`. The 3 pre-existing tests still pass; one assertion changed because `sourceFile` now names the account (`keychain:Claude Code-credentials (<user>)`), which is display-only — it feeds a tooltip in `providers.tsx`.
- `npm run test:core`: **651/660**. The 4 failures (`RequestLogStore bounded heap regression`, 3 Claude Design plugin config) are identical on an unmodified tree at `v3.0.17` (645/654), so they are pre-existing and unrelated.
- `tsc --noEmit` clean.
- End-to-end on a real machine with Claude Code 2.1.220 and a subscription login: `getLocalAgentProviderCandidates` over the live web RPC returns `status=available importable=true`; import produces a provider whose bearer gets **HTTP 200** from `GET /api/oauth/usage` with real quota data, and a real inference request through the gateway on `/v1/messages` returns **200**. So it resolves the correct live token, not an MCP one.

Tested only on macOS/arm64 — the changed code paths are all `process.platform === "darwin"`-gated.

### DIFF
--- a/packages/core/src/agents/local-providers/claude-code.ts
+++ b/packages/core/src/agents/local-providers/claude-code.ts
@@ -1,4 +1,5 @@
 import { execFileSync } from "node:child_process";
+import { createHash } from "node:crypto";
 import os from "node:os";
 import path from "node:path";
 import type {
@@ -21,7 +22,23 @@ import {
 } from "@ccr/core/agents/local-providers/shared";
 
 const claudeDefaultModels = ["claude-sonnet-5"];
-const claudeCodeKeychainService = "Claude Code-credentials";
+const claudeCodeKeychainServiceBase = "Claude Code-credentials";
+const claudeCodeKeychainAccountPattern = /^[a-zA-Z0-9._-]+$/;
+const claudeCodeKeychainServicePattern = /^Claude Code(?:-[a-z]+-oauth)?-credentials(?:-[0-9a-f]{8})?$/;
+// `security` exit code for errSecItemNotFound.
+const keychainItemNotFoundStatus = 44;
+
+type ClaudeCodeKeychainCandidate = { account?: string; service: string };
+
+export type ClaudeCodeLoginScan = {
+  /** Reads that failed, with the `security` stderr that explains why. */
+  errors: string[];
+  /** Locations that parsed successfully. */
+  inspected: string[];
+  oauth?: OAuthTokenSet;
+  /** Locations that parsed but carry no OAuth token. */
+  tokenless: string[];
+};
 
 const percentLimitMapping = (id: string, label: string, path: string, window: string) => ({
   id,
@@ -65,7 +82,8 @@ const claudeCodeAccountMapping: ProviderAccountMappingConfig = {
 };
 
 export function claudeCodeCandidate(): LocalAgentProviderCandidate {
-  const oauth = readClaudeCodeOauth();
+  const scan = scanClaudeCodeLogin();
+  const oauth = scan.oauth;
   if (oauth?.accessToken) {
     return {
       detail: "Claude Code login detected. Click Import to add it as a gateway provider.",
@@ -92,7 +110,34 @@ export function claudeCodeCandidate(): LocalAgentProviderCandidate {
       status: "locked"
     };
   }
+  const detail = claudeCodeScanDiagnostic(scan);
+  if (detail) {
+    return {
+      detail,
+      id: "claude-code-api",
+      importable: false,
+      kind: "claude-code",
+      models: claudeDefaultModels,
+      name: "Claude Code API",
+      protocol: "anthropic_messages",
+      sourceFile: scan.inspected[0],
+      status: "locked"
+    };
+  }
   return missingCandidate("claude-code", "claude-code-api", "Claude Code API", "anthropic_messages", claudeDefaultModels);
+}
+
+// Distinguishes "found login state but no token" and "could not read the store"
+// from "no login at all", so the Add Provider list can say why an import is
+// unavailable instead of silently omitting the candidate.
+function claudeCodeScanDiagnostic(scan: ClaudeCodeLoginScan): string | undefined {
+  if (scan.tokenless.length > 0) {
+    return `Claude Code login state was found but contains no OAuth token (${scan.tokenless.join("; ")}). Run \`claude /login\`, then retry.`;
+  }
+  if (scan.errors.length > 0) {
+    return `Claude Code login state could not be read: ${scan.errors.join("; ")}`;
+  }
+  return undefined;
 }
 
 export function importClaudeCodeProvider(candidate: LocalAgentProviderCandidate, providerNames: string[]): LocalAgentProviderImportResult {
@@ -139,9 +184,15 @@ function claudeCodeProviderAccountConfig(): ProviderAccountConfig {
 }
 
 export function readClaudeCodeOauth(): OAuthTokenSet | undefined {
-  const keychainOauth = readClaudeCodeKeychainOauth();
+  return scanClaudeCodeLogin().oauth;
+}
+
+export function scanClaudeCodeLogin(): ClaudeCodeLoginScan {
+  const scan: ClaudeCodeLoginScan = { errors: [], inspected: [], tokenless: [] };
+  const keychainOauth = scanClaudeCodeKeychain(scan);
   if (keychainOauth) {
-    return keychainOauth;
+    scan.oauth = keychainOauth;
+    return scan;
   }
 
   for (const sourceFile of claudeCredentialFiles()) {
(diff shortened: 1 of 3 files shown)


musistudio:
Thanks for the detailed fix. The account-specific Keychain lookup looks correct and the new coverage around `$USER`, service suffixes, diagnostics, and logged-out behavior is solid.

I found one remaining blocker around the `mcpOAuth` fallback. The PR correctly prefers `claudeAiOauth` when it exists, but if a Keychain record only contains `mcpOAuth` entries with `accessToken`, the recursive fallback still imports that plugin token as a Claude Code provider. That recreates the 401 failure mode the PR is trying to avoid.

I think the fallback should not recursively search through `mcpOAuth`. It should either only accept explicit `claudeAiOauth`, or restrict legacy fallback to root-level Claude token fields.

The tests below fail on the current PR and capture the gap.

```javascript
test("Claude Code local provider does not import an mcpOAuth-only keychain token", { skip: process.platform === "win32" }, async () => {
  await withClaudeCodeHome(async () => {
    await withPlatform("darwin", async () => {
      await withFakeSecurityKeychain([
        {
          account: keychainAccount,
          modified: "20260729042837",
          record: {
            mcpOAuth: {
              "plugin:github|1": {
                accessToken: "mcp-plugin-token",
                refreshToken: "mcp-plugin-refresh"
              }
            }
          },
          service: "Claude Code-credentials"
        }
      ], async () => {
        const candidate = claudeCodeCandidate();

        assert.equal(candidate.status, "locked");
        assert.equal(candidate.importable, false);
        assert.match(candidate.detail, /no OAuth token/);
        assert.match(candidate.detail, /mcpOAuth/);
        assert.throws(
          () => importClaudeCodeProvider(candidate, []),
          /Claude Code access token was not found/
        );
      });
    });
  });
});

test("Claude Code local provider prefers file credentials over an mcpOAuth-only keychain token", { skip: process.platform === "win32" }, async () => {
  await withClaudeCodeHome(async (home) => {
    await withPlatform("darwin", async () => {
      await withFakeSecurityKeychain([
        {
          account: keychainAccount,
          modified: "20260729042837",
          record: {
            mcpOAuth: {
              "plugin:github|1": {
                accessToken: "mcp-plugin-token",
                refreshToken: "mcp-plugin-refresh"
              }
            }
          },
          service: "Claude Code-credentials"
        }
      ], async () => {
        const credentialFile = writeClaudeCredentials(home, {
          accessToken: "file-access-token",
          refreshToken: "file-refresh-token"
        });

        const candidate = claudeCodeCandidate();

        assert.equal(candidate.status, "available");
        assert.equal(candidate.importable, true);
        assert.equal(candidate.sourceFile, credentialFile);

        const result = importClaudeCodeProvider(candidate, []);
        assert.equal(result.providerPlugins[0].auth.headers.authorization, "Bearer file-access-token");
      });
    });
  });
});
```

matteoantoci:
You're right, and thanks for the concrete tests — they made the gap unambiguous. Fixed in 70d31ab.

I went with your **second** option, restricting the fallback to root-level token fields, rather than accepting only `claudeAiOauth`. Two reasons:

- Root-level excludes *any* nested container, not just `mcpOAuth`. Special-casing the key name by exclusion (e.g. searching `{...record, mcpOAuth: undefined}`) leaves the same hole open for a future sibling key.
- `claudeAiOauth`-only would break the existing `Core gateway config replaces imported Claude Code OAuth token with live macOS Keychain token` test, whose fixture is a root-level `{ access_token, refresh_token }` record. The pre-`claudeAiOauth` layout kept tokens at the record root, so that path is worth keeping.

So both reads are now non-recursive:

```ts
const claudeAiOauth = readOauthTokenSetFields(record.claudeAiOauth);
if (claudeAiOauth) { /* ... */ }
// Pre-`claudeAiOauth` layout, where the tokens sat at the record root.
const legacyRoot = readOauthTokenSetFields(record);
```

`readOauthTokenSetFields()` is extracted out of `findOauthTokenSet()` in `shared.ts` rather than duplicating the field list; `findOauthTokenSet()` keeps its existing behaviour, so codex/grok/kimi/opencode/zcode are untouched.

I also applied the same restriction to the **credentials-file** path, which had the identical hazard — Claude Code writes `.credentials.json` with the same record shape, `mcpOAuth` included, so a recursive search there could pick a plugin token too.

Your two tests are in verbatim. Confirmed red on the commit you reviewed (519b92d) and green now — reverting only the two `src/` files with the tests in place fails exactly those two and nothing else:

```
✖ does not import an mcpOAuth-only keychain token
✖ prefers file credentials over an mcpOAuth-only keychain token
ℹ tests 11 | pass 9 | fail 2      <- 519b92d
ℹ tests 11 | pass 11 | fail 0     <- 70d31ab
```

`npm run test:core` is 653/662; the 4 failures are the same pre-existing ones (`RequestLogStore bounded heap regression`, 3 Claude Design plugin config), identical on an unmodified `v3.0.17` tree. `tsc --noEmit` clean. Re-verified end to end on the live machine: candidate resolves `available` from the keychain, and a real request through the gateway on `/v1/messages` returns 200.

---

## Separate finding on #1567, out of scope here

While re-verifying I hit a genuine 401 that has nothing to do with which item gets read, and I think it's the actual #1567:

```
keychain claudeAiOauth.expiresAt  2026-07-29T12:28:37.784Z
now                              2026-07-29T12:29:31.398Z   -> expired
GET /v1/messages with that token -> 401 "OAuth access token has expired."
```

`claudeAiOauth` carries `refreshToken`, `expiresAt` and `refreshTokenExpiresAt` right next to `accessToken`, and CCR uses none of them — so an import goes stale roughly hourly and there is no refresh path.

There's a second, sharper edge. After Claude Code refreshed its own token, the still-running gateway returned:

```
401 "OAuth access token has been revoked."
```

and only recovered after `launchctl kickstart`. The bearer is snapshotted into `providerPlugins` at config-compile time, which happens at startup, so a long-lived gateway keeps presenting a token that Claude Code has already rotated and Anthropic has revoked. That matches the #1567 report exactly: import detected, works briefly, then 401s.

Happy to do that as a follow-up PR — refresh via `POST /v1/oauth/token` with the stored `refreshToken` when `expiresAt` has passed, plus a re-read on upstream 401. It needs care around concurrent refreshes: Claude Code does a compare-and-swap on `refreshToken` when it writes, so a naive out-of-band refresh from CCR would rotate the token out from under the CLI and force a re-login. Probably wants the same lock discipline, or a read-only "re-read the keychain on 401" that lets the CLI own refreshing. Let me know which you'd prefer and I'll open it separately.

musistudio:
Looks good to me. I'll merge it now.

PR #1606, opened 2026-07-29
fix: validator thinking blocks + SSE streaming termination (npm 2.1.1 dist)

im47cn:
# Fix: Anthropic validator + SSE streaming termination (npm 2.1.1 dist)

## Problem 1: Validator rejects thinking/redacted_thinking blocks

Claude Code CLI 2.1.x introduced **interleaved-thinking** and **redact-thinking** features. When CCR receives these requests, the Anthropic content block validator throws `Unknown content block type "thinking"` / `"redacted_thinking"`, causing all requests with thinking blocks to fail.

## Problem 2: SSE streaming never terminates — Claude Code repeats output

The npm 2.1.1 dist's `handleStreamingRequest` has three compounding bugs:

1. **Duplicate framing events**: Provider's `message_start`, `ping`, `message_delta`, `message_stop` are transparently forwarded alongside the handler's own copies. Claude Code sees double events per turn.

2. **`stop_reason` deleted**: `message_delta`'s `stop_reason` is explicitly stripped (code comment: "allow conversation continuation"). Claude Code never receives the `end_turn` signal.

3. **`message_stop` filtered**: The `message_stop` event is explicitly suppressed. Claude Code cannot determine that a response turn has completed.

**Combined effect**: Claude Code retries indefinitely, producing repeated output when using GLM or other non-Anthropic providers via CCR.

## Root Cause

Both bugs exist only in the npm-published `dist/`. The TypeScript source in `feature/2.0` already handles both correctly:
- `anthropic.transformer.ts`: thinking blocks supported (lines 163-171)
- `anthropic.transformer.ts` `convertOpenAIStreamToAnthropic()`: stop_reason and message_stop correctly emitted

The npm `dist/` is published from a different commit (orphan code).

### Source (Problem 1 only)
1. Added `packages/core/src/input/anthropic/validator.ts` with full content block support including thinking blocks.
2. Added cases for `thinking`, `redacted_thinking`, `server_tool_use`, `web_search_tool_result`, `container_upload`.
3. Changed `default` to `break` instead of throwing (forward-compatible).
4. Added 20 unit tests.

### Bridge script (Problems 1 + 2)
Added `scripts/patch-npm-dist.sh` — patches the bundled dist in-place after `npm update`:
- Validator block-type fix (cli.js + validator.js)
- SSE streaming: filter provider framing events + restore stop_reason/message_stop

## Testing

```bash
cd packages/core
pnpm test -- validator   # 20/20 pass
```

SSE verification (after patching dist):
```
message_start -> ping -> content_block_start -> content_block_delta(s) -> content_block_stop -> message_delta -> message_stop
```

### DIFF
--- a/.pr-description.md
+++ b/.pr-description.md
@@ -1,0 +1,33 @@
+# Fix: Anthropic content block validator rejects thinking/redacted_thinking blocks
+
+## Problem
+
+Claude Code CLI 2.1.x introduced **interleaved-thinking** and **redact-thinking** features. When CCR receives these requests, the Anthropic content block validator throws `Unknown content block type "thinking"` / `"redacted_thinking"`, causing all requests with thinking blocks to fail.
+
+This affects:
+- `claude --model claude-sonnet-4-20250514` with extended thinking enabled
+- Any client sending `thinking` or `redacted_thinking` content blocks per the [Anthropic Messages API](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking)
+
+## Root Cause
+
+The content block `switch` statement in `validateContentBlocks()` had no cases for `thinking`, `redacted_thinking`, `server_tool_use`, `web_search_tool_result`, or `container_upload`. The `default` branch threw a `ValidationError` for any unrecognized type.
+
+## Solution
+
+1. **Added `packages/core/src/input/anthropic/validator.ts`** — a proper TypeScript validator source with full content block support including thinking blocks. This source was missing from the repo (it only existed in the npm-published `dist/`).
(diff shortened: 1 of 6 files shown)


TonyGeez:
Hello, i'm confused about your mention of "missing" dist/ in a git repo, can you elaborate please.

Also, Anthropic API accepts string or array but your code reject string. This will most likely cause error on production.

im47cn:
Closing this PR — both bugs (validator thinking blocks + SSE streaming termination) are already fixed in v3.0.17 (@musistudio/claude-code-router). The old npm package `claude-code-router` (published by fanzhang16, repo now 404) is deprecated. Users should upgrade to v3.x.

Thanks for the review time.

PR #1618, opened 2026-07-31
fix(ui): repair provider setup flow in Get Started onboarding

### LINKED ISSUE #1617: [Bug] New UI Get Started Onboarding wizard does not trigger to load list models after adding provider
It breaks from version 3.0.16+ (even lastest 3.0.18 still facing the issue). Rollback to 3.0.15 with the old UI in adding models is working
There is no way to trigger it loads here (even try Back to API key input and Next step again).

Image

VietND96:
Fixes #1617 

Here is result of local build test
image (1)

In next step, dialog to select model tick to the dropdown (before that, need to maximize the window to see it displays in the bottom)
image (2)

The onboarding wizard and the Add/Edit Provider dialog render the same AddProviderForm, but the wizard is wired up differently, and several defects lived in those differences.

Model probe never fired during onboarding. The model-probe useEffect in App.tsx guarded on `providerAddOpen` only, which is set true solely when opening the manual Add/Edit Provider dialog. It is never set during the onboarding wizard, so after entering an endpoint + API key on the Get Started flow and reaching the "Pick models" step,
probeProviderCandidates() never fired and the model list stayed empty with no loading state. Regressed in 9f704fc ("Guide onboarding through granular provider setup steps"), which narrowed the guard from `providerAddOpen || (activeView === "onboarding" && onboardingStep === "provider")` down to `providerAddOpen`. Restore the broader guard so the probe runs both when the dialog is open and when onboarding is on the provider step.

A rejected API key produced an empty model list with no feedback. Typing an API key switches the probe from mode "protocols" to mode "models", but the error-reporting branch was guarded on `probeMode !== "models"`, so it never ran in exactly that mode. The backend still runs full protocol probing in models mode, so a 401 came back as `supported: false` with a real message that was then discarded. Report the failure, while staying quiet when models were discovered, since a provider can expose a working catalog while a protocol probe endpoint 404s.

"Added models" faked a loading state for data it already had. The panel renders local draft state, yet showed a skeleton and hid its whole toolbar whenever the catalog probe re-ran, leaving the header badge reading a real count above shimmer rows and removing the "Custom model" button, the only way to add models on a provider with no catalog. Keep both panels' controls mounted, disabling the catalog search while loading, which also removes a layout shift and mid-typing focus loss.

An empty catalog gave no way forward, so point at "Custom model" (en+zh).

Onboarding's "Check Connection" spent real credits with no confirmation. The dialog wraps onCheck in a confirm step that warns about account balance, lets the user pick models, and shows per-model results; onboarding passed onCheckProvider straight through, so one click fired a billable request against every configured model and discarded the report. Extract that step as ProviderConnectivityCheckDialog and use it from both surfaces.

ProviderConnectionStatusRow hardcoded bg-emerald-50 / border-emerald-200 / bg-amber-50 with no dark variant, so the "Verify connection" status chips rendered as bright light blobs on the dark card. Use alpha fills, matching the pattern used elsewhere in the file.

The provider error banner was not announced; add role="alert".


VietND96:
Many people in our org confirmed the same issue when getting started with the latest version. Could you please review @musistudio? Thanks

Issue #13, opened 2025-06-09
can ROUTER_AGENT_MODEL, TOOL_AGENT_MODEL be deepseek model?


tblinka1996:
Hello

I want to use ROUTER_AGENT_MODEL, TOOL_AGENT_MODEL with deepseek third part API URL such as nebius.
Will it support for above to ROUTER_AGENT_MODEL, TOOL_AGENT_MODEL?

musistudio:
I’ll release a new version later, and in the new version, you won’t need to set these variables anymore. The deepseek-v3 model used by claude code amazed me in both performance and cost.

musistudio:
I’ve released a new version, but it’s not very stable yet (some issues haven’t been fully handled). You can try it out using the DeepSeek API, and feel free to share any feedback or issues!

tblinka1996:
thanks, it works for me now :-)

PR #1625, opened 2026-08-02
feat(providers): support model-level protocol features

mindops:
## Background

DeepSeek-compatible Responses APIs may return plaintext reasoning items, while OpenAI Responses expects encrypted reasoning state or no raw reasoning history. When a conversation is switched from a DeepSeek-compatible upstream to OpenAI, forwarding plaintext reasoning history can make OpenAI reject the request.

Error example:

Error screenshot

~~~text
Invalid 'input[7].content': array too long. Expected an array with maximum length 0, but got an array with length 1 instead.
~~~

Provider-level protocol detection is also not precise enough. A provider can expose multiple OpenAI-compatible protocols, while individual models may only support a subset or require different Responses reasoning handling.

## Before

Previously, provider config could describe provider-level protocols/capabilities, but it could not describe protocol-specific behavior or model-level protocol support.

~~~json
{
  "Providers": [
    {
      "name": "DeepSeek Compatible",
      "type": "openai_responses",
      "api_base_url": "https://api.deepseek.com",
      "models": ["deepseek-reasoner", "custom-gpt-5.5"],
      "capabilities": [
        {
          "type": "openai_responses",
          "baseUrl": "https://api.deepseek.com"
        },
        {
          "type": "anthropic_messages",
          "baseUrl": "https://api.deepseek.com"
        }
      ]
    }
  ]
}
~~~

This had two gaps:

- There was no way to say that this Responses-compatible provider returns plaintext reasoning history.
- There was no way to say that `custom-gpt-5.5` supports only a subset of the provider protocols.

## After

Provider-level protocol features can now define the default behavior for a protocol.

~~~json
{
  "Providers": [
    {
      "name": "DeepSeek Compatible",
      "type": "openai_responses",
      "api_base_url": "https://api.deepseek.com",
      "models": ["deepseek-reasoner", "custom-gpt-5.5"],
      "capabilities": [
        {
          "type": "openai_responses",
          "baseUrl": "https://api.deepseek.com",
          "features": {
            "reasoning_history_policy": "plaintext",
            "reasoning_summary_policy": "as_content"
          }
        },
        {
          "type": "anthropic_messages",
          "baseUrl": "https://api.deepseek.com"
        }
      ]
    }
  ]
}
~~~

Model-level metadata can further narrow supported protocols or override protocol features.

~~~json
{
  "Providers": [
    {
      "name": "Mixed Compatible Provider",
      "type": "openai_responses",
      "api_base_url": "https://gateway.example/v1",
      "models": ["deepseek-reasoner", "custom-gpt-5.5"],
      "capabilities": [
        {
          "type": "openai_responses",
          "baseUrl": "https://gateway.example/v1",
          "features": {
            "reasoning_history_policy": "strip"
          }
        },
        {
          "type": "anthropic_messages",
          "baseUrl": "https://gateway.example/v1"
        }
      ],
      "modelMetadata": {
        "deepseek-reasoner": {
          "protocols": ["openai_responses"],
          "protocolFeatures": {
            "openai_responses": {
              "reasoningHistoryPolicy": "plaintext",
              "reasoningSummaryPolicy": "as_content"
            }
          }
        },
        "custom-gpt-5.5": {
          "protocols": ["openai_responses"]
        }
      }
    }
  ]
}
~~~

Provider-level UI:

Provider-level config

Model-level UI:

Model-level config

## Changes

- Centralize gateway protocol definitions into shared runtime constants and type guards.
- Add provider capability `features` for protocol-specific behavior.
- Add model metadata `protocols` and `protocolFeatures`.
- Make upstream protocol selection model-aware.
- Sanitize OpenAI Responses reasoning history before forwarding upstream:
  - `encrypted`: keep encrypted reasoning state, strip plaintext content.
  - `plaintext`: keep plaintext reasoning content, strip encrypted state.
  - `strip`: remove reasoning history.
- Add `reasoning_summary_policy` for converting summary into reasoning content when needed.
- Add provider/model UI controls for protocol support and Responses compatibility.
- Fix checkbox label binding in provider/model protocol settings.

## Design

The config is intentionally split into two levels:

- Provider level defines the protocols and default protocol behavior exposed by the upstream.
- Model level optionally narrows or overrides the provider defaults for a specific model.

Resolution order:

~~~text
model protocolFeatures
> provider capability features
> baseUrl auto inference
~~~

For Responses reasoning history, auto inference only uses `baseUrl`:

- `api.openai.com` -> `encrypted`
- `api.deepseek.com` -> `plaintext`
- unknown Responses-compatible upstream -> `strip`

Model protocol inheritance rules:

- Missing `modelMetadata[model].protocols` means inherit provider protocols.
- Explicit `protocols: []` means the model supports no protocols.
- Model protocols are always limited to protocols supported by the provider.

## Extensibility

This structure keeps future protocol work localized.

To add another gateway protocol:

1. Add it to the shared gateway protocol list.
2. Add its UI label.
3. Add probe/routing behavior if needed.
4. Add protocol-specific features only if the protocol has compatibility differences.

To add another protocol feature:

1. Extend provider capability `features`.
2. Optionally allow model-level override through `protocolFeatures`.
3. Apply the feature in the relevant upstream normalization/routing path.
4. Add UI controls only when the feature needs to be user-configurable.

This avoids adding one-off provider/model checks for every compatible API variant.

## Testing

- `npm run typecheck`
- `git diff --check`
- `node build/test.mjs core --scope unit`
- `npm --workspace packages/ui run test:integration`


musistudio:
I think this input relies on upstream dependency https://github.com/The-NeXT-AI/ai-gateway
There are relevant PRs available https://github.com/The-NeXT-AI/ai-gateway/pull/13

mindops:
Thanks, that makes sense.

I looked into the current dependency path and The-NeXT-AI/ai-gateway#13. I agree that the actual reasoning/thinking transformation and encrypted-state replay should live in ai-gateway rather than CCR duplicating that logic.

My current understanding is that this PR mixes two layers:

1. CCR-owned config/routing/UI:
   - provider-level protocol support
   - model-level protocol support
   - model-aware protocol selection
   - UI for provider/model protocol configuration

2. Gateway-owned request normalization:
   - stripping or preserving Responses reasoning history
   - deciding whether plaintext reasoning, `encrypted_content`, or summary should be forwarded
   - converting summary into reasoning content when needed

Layer 2 overlaps with ai-gateway. PR #13 adds the right lower-level direction there: separate readable reasoning text/summary from opaque encrypted state, and only replay encrypted state when the source origin matches.

The gap I found is that ai-gateway currently has compatibility controls such as `openaiChatReasoningSplit`, `openaiChatThinkingOptions`, and the DeepSeek thinking plugin, but it does not yet expose a configurable Responses reasoning-history policy equivalent to:

- encrypted / plaintext / strip
- drop summary / summary as content
- provider default with model-level override

So I think the cleaner split is:

- CCR keeps provider/model protocol metadata, model-aware routing, and UI.
- ai-gateway owns reasoning transport, conversion, and encrypted-state replay safety.
- CCR resolves provider/model policy and passes it into ai-gateway once ai-gateway exposes this config surface.

I can rework this PR to keep the CCR-side protocol configuration and remove/minimize the local Responses reasoning sanitizer, then follow up in ai-gateway with the missing policy config or wait for #13 to land and build on top of it.

PR #1627, opened 2026-08-03
feat(ui): polish account balance cards and fix connector persistence

jaypume:
## Summary

Polish the account balance cards across the dashboard and tray, and fix two bugs that silently discarded custom HTTP JSON account connectors.

### Account balance UI
- Meter rows are now compact single lines: label, reset countdown (`🕛 3h28m` / `☀️ 1d17h`) with monospace digits, percentage right-aligned.
- Day-scale windows (7D/30D) use `☀️`, hour-scale (5H) uses `🕛`; the wordy `expires in` prefix is dropped and durations are zero-padded (`3h28m`, `1d17h`) so rows align.
- Percent values warn in amber at ≥75 and turn red at ≥90 (percent-unit meters only).
- The tray account panel and the dashboard gauge/label surfaces share the same compact format (`formatProviderAccountMeterTitleCompact`) with monospace digits.

### Fixes: account connector persistence
- Saving with `accountMode: "raw"` no longer falls into the HTTP JSON form branch when a stale form URL is present; pasted Connectors JSON is now always used for raw mode.
- Reopening an http-json connector whose meters are quota/tokens/time-window keeps them intact in raw mode instead of flattening everything into a single `subscription` meter (which dropped meters and the window).
- Regression tests cover both paths.

## Verification
- `tsc --noEmit`: clean.
- ui workspace tests: 152 passing (component + integration).
- Manual: configured a MiniMax coding-plan http-json connector (5H用量 / 7D用量 meters), confirmed it saves, reopens in raw mode, and renders the compact format in both dashboard and tray.

### DIFF
--- a/packages/ui/src/pages/home/components/dashboard.tsx
+++ b/packages/ui/src/pages/home/components/dashboard.tsx
@@ -7,7 +7,7 @@ import {
   compactUserAgent, compareProviderAccountSnapshots, ComposedChart, CSS, DEFAULT_OVERVIEW_WIDGETS, DndContext,
   Dialog, DialogBody, DialogContent, DialogHeader, DialogTitle,
   DragEndEvent, DragOverEvent, DragOverlay, DragStartEvent, Field, formatAxisNumber, formatBytes,
-  formatCompactNumber, formatDuration, formatLogDateTime, formatPercent, formatProviderAccountDetailDate, formatProviderAccountMeterTitle, formatProviderAccountMeterValue,
+  formatCompactNumber, formatDuration, formatLogDateTime, formatPercent, formatProviderAccountDetailDate, formatProviderAccountMeterTitle, formatProviderAccountMeterValue, formatProviderAccountResetDuration,
   formatStatusBucketDate, formatSystemStatusRange, formatUsdCost, KeyboardSensor,
   LabelList, LayoutGroup, Line, LoaderCircle, MeasuringStrategy, MetricTone,
   motion, normalizeAgentFilterValue, normalizeOverviewWidget, normalizeOverviewWidgets,
@@ -2347,7 +2347,7 @@ function ProviderAccountsOverview({
             })}
           </div>
         ) : variant === "bars" ? (
-          <div className={cn("h-full min-h-0 overflow-y-auto pr-1", providerAccountStackClass(dimensions))}>
+          <div className={cn("h-full min-h-0 overflow-y-auto pr-1", "space-y-1.5")}>
             {visibleAccounts.map((account) => {
               const meter = primaryProviderAccountDisplayMeter(account);
               const progress = meter && isProviderAccountQuotaMeter(meter) ? providerAccountMeterProgress(meter) : undefined;
@@ -2422,7 +2422,7 @@ function ProviderAccountSinglePanel({
   const showQuotaVisual = providerAccountUsesQuotaVisual(variant) && quotaMeters.length > 0;
 
   return (
-    <div className={cn("flex h-full min-h-0 min-w-0 flex-col overflow-hidden", providerAccountStackClass(dimensions))}>
+    <div className={cn("flex h-full min-h-0 min-w-0 flex-col overflow-hidden", "space-y-1.5")}>
       <div className="flex min-w-0 shrink-0 items-start justify-between gap-3">
         <div className="min-w-0">
           <div className={cn("truncate font-semibold", dimensions.height <= 1 ? "text-[12px]" : "text-[13px]")}>{providerAccountSnapshotLabel(account)}</div>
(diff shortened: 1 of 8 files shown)


packages/ui/src/pages/home/shared/providers.ts
```diff
@@ -1087,7 +1087,9 @@ export function parseProviderAccountDraft(draft: AddProviderDraft): GatewayProvi
     };
   }
 
-  if (draft.accountMode === "http-json" || draft.usageRequestUrl.trim()) {
+  // raw 模式必须走 Connectors JSON；若用 ||，表单 URL 残留会让 raw 模式
+  // 误走表单生成分支，用户粘贴的 connectors 被静默丢弃（issue: 5H用量/7D用量 变 Subscription）
+  if (draft.accountMode === "http-json" && draft.usageRequestUrl.trim()) {
```
musistudio, line 1092: This should still enter `providerHttpJsonConnectorFromDraft` whenever `accountMode` is `"http-json"`, even when the URL is empty. With the new `&& draft.usageRequestUrl.trim()` guard, an empty URL skips the HTTP JSON form validation and falls through to the raw connectors JSON branch, which the UI does not show in this mode. That can surface the wrong error, or save stale raw connectors if `accountConnectorsText` happens to contain valid JSON. A safer condition is `draft.accountMode === "http-json"`; raw mode is already excluded.

packages/ui/src/pages/home/components/dashboard.tsx
```diff
@@ -2556,10 +2567,20 @@ function ProviderAccountMeterLine({
...
   const [resetDialogDetail, setResetDialogDetail] = useState<NonNullable<ProviderAccountMeter["details"]>[number]>();
-  const title = formatProviderAccountMeterTitle(meter, t);
+  const labelText = t(meter.label);
+  const resetDuration = meter.resetAt ? formatProviderAccountResetDuration(meter.resetAt) : undefined;
+  // 已过期（或非法时间）时回退为「expired」文案；正常时用图标 + 纯时长，不重复「剩余」前缀。
+  // 天级窗口（如 7D/30D，时长含 d 单位）用 ☀️，小时级（如 5H）用 🕛。
+  const resetIcon = resetDuration?.includes("d") ? "☀️" : "🕛";
+  const resetText = resetDuration !== undefined ? `${resetIcon} ${resetDuration}` : (meter.resetAt ? t("expired") : "");
+  // 百分比告警分级：达到 75 用橙色（text-amber-500），达到 90 及以上用红色（text-red-500）
+  const meterNumber = meter.remaining ?? meter.used ?? meter.limit;
```
musistudio, line 2577: This threshold is using the displayed percent, but most percent quota meters in this codebase display `remaining` first. For example the built-in Zhipu/Zai/Claude Code mappings use `remaining: 100 - utilization` and `used: utilization`, and the account status logic marks low remaining ratio as warning/critical. With this code, `96%` remaining is colored red even though it is healthy. If the intent is usage warning colors, prefer `meter.used`; if the displayed value is remaining, the thresholds need to be inverted.

packages/ui/src/pages/home/shared/providers.ts
```diff
@@ -1193,9 +1195,22 @@ export function createProviderAccountDraftFromConfig(account: ProviderAccountCon
   }
 
   const balanceMeter = httpJsonConnector.mapping.meters.find((meter) => meter.kind === "balance" || meter.id === "balance");
-  const subscriptionMeter = httpJsonConnector.mapping.meters.find((meter) =>
-    meter.kind === "subscription" || meter.id === "subscription" || meter.kind === "quota" || meter.kind === "tokens" || meter.kind === "time_window"
+  const subscriptionMeter = httpJsonConnector.mapping.meters.find((meter) => meter.kind === "subscription" || meter.id === "subscription");
+  // 仅当所有 meter 都是简单的 balance/subscription 时才展开为表单字段；
+  // 含 quota/tokens/time_window 或自定义结构的 meter（如 MiniMax 的 5H用量/7D用量）
+  // 展开会丢 meter 和 window，必须保留 raw Connectors JSON。
+  const simpleHttpJsonConnector = httpJsonConnector.mapping.meters.every((meter) =>
```
musistudio, line 1202: This simple connector check is still broader than what the HTTP JSON form can losslessly round-trip. The form can represent at most one balance meter and one subscription meter, and saving rebuilds connector-level fields such as `auth`. A connector with multiple balance/subscription meters or non-default connector metadata can still be flattened after reopen-save. Consider only expanding connectors that exactly match the form representable shape; otherwise keep raw mode.

musistudio (changes requested):
Thanks for the connector persistence fixes and the compact account UI cleanup. I found a couple of regressions that should be fixed before this lands: HTTP JSON mode can now skip its own validation when the URL is empty, and the new percent coloring treats high remaining quota as dangerous. I also left a smaller round-trip concern on the simple HTTP JSON connector detection.

PR #1656, opened 2026-08-11
fix: expose core gateway auth token before the HTTP server binds

STRML:
## What

The gateway service assigned `coreAuthToken` after `listen()` had already started accepting requests. A request landing in that startup window reached the executor with an empty token and returned a misleading `Core gateway auth token is not initialized.` 502.

## Fix

Set `coreAuthToken` before `listen()` binds. The token is still cleared on the failure paths (`stop()`, `handleCoreGatewayTermination`), so a failed child spawn or a terminated child behaves exactly as before - only the startup-window requests change.

## Why it matters

The 502 is confusing: it points at the auth token when the real cause is a race between `listen()` and the managed gateway child spawn. On a busy machine (or right after a config reload) the window is wide enough that real requests hit it, and the error sends you down the wrong debugging path.

## Test

`gateway startup never serves a request with an uninitialized core auth token` - starts the gateway un-awaited, fires 8 concurrent requests the moment the port accepts, and asserts none returns the misleading error. Fails on the old ordering, passes with the fix.

Verified: `tsc --noEmit` clean, the new test passes, and the core suite's pre-existing failures are unchanged.

### DIFF
--- a/packages/core/src/gateway/application/gateway-service.ts
+++ b/packages/core/src/gateway/application/gateway-service.ts
@@ -159,6 +159,13 @@ class GatewayService {
         return this.status;
       }
 
+      // The HTTP server starts accepting requests as soon as listen() binds, but
+      // the managed core gateway child is spawned later in this method. Expose
+      // the auth token before binding so a request that lands in that startup
+      // window is not rejected with a misleading "core gateway auth token is not
+      // initialized" 502. The token is cleared again by stop() if the child
+      // fails to spawn, and by handleCoreGatewayTermination if it exits later.
+      this.coreAuthToken = coreAuthToken;
       await this.rawTraceSynchronizer.start();
       await this.listen(config);
       if (this.server) {
@@ -186,7 +193,6 @@ class GatewayService {
         const runtimeId = randomUUID();
         const spawnedGateway = spawnGatewayProcess(config, coreGatewayConfig, upstreamProxyUrl, runtimeId, coreAuthToken);
         this.child = spawnedGateway.child;
-        this.coreAuthToken = coreAuthToken;
         const managedChild = this.child;
         let markerWritePromise: Promise<void> | undefined;
         let startupFailure: Error | undefined;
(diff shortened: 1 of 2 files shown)


wanghongqu:
+1

pacocartones (commented):
Read through this end to end — the fix looks correct to me.

The one thing I checked closely was whether moving `this.coreAuthToken = coreAuthToken`
earlier could expose a not-yet-finalized token: `coreAuthToken` is assigned from
`generateCoreGatewayAuthToken()` at the very top of `start()` (before `mediaService.start()`,
`pluginService.start()`, and everything else in the method), so by the time it reaches the new
assignment point right before `listen()`, it's already a complete, valid token — there's no
partial-construction window to worry about.

The failure-path cleanup also stays correct: `stop()` still clears `coreAuthToken` back to `""`
unconditionally, and `handleCoreGatewayTermination` still clears it on child exit, so a failed
spawn or an early termination behaves exactly as before — only the startup-window requests
change, as the PR description says.

The test is a good one: firing 8 concurrent requests the moment the port accepts and asserting
none hits the misleading auth-token error is a real regression guard for a race condition, not
just a happy-path check.

TonyGeez:
> Read through this end to end — the fix looks correct to me.
> 
> The one thing I checked closely was whether moving `this.coreAuthToken = coreAuthToken` earlier could expose a not-yet-finalized token: `coreAuthToken` is assigned from `generateCoreGatewayAuthToken()` at the very top of `start()` (before `mediaService.start()`, `pluginService.start()`, and everything else in the method), so by the time it reaches the new assignment point right before `listen()`, it's already a complete, valid token — there's no partial-construction window to worry about.
> 
> The failure-path cleanup also stays correct: `stop()` still clears `coreAuthToken` back to `""` unconditionally, and `handleCoreGatewayTermination` still clears it on child exit, so a failed spawn or an early termination behaves exactly as before — only the startup-window requests change, as the PR description says.
> 
> The test is a good one: firing 8 concurrent requests the moment the port accepts and asserting none hits the misleading auth-token error is a real regression guard for a race condition, not just a happy-path check.

There's no need to ask AI to review the commit of a PR just to copy/paste its response into a PR comment. I don't think this will influence anything in any way.

PR #1680, opened 2026-08-17
fix(ui): keep hand-written provider fields when saving from the dialog

songkuan-zheng:
## Problem

Editing a provider in the web UI silently deletes every provider field the setup dialog does not render.

`saveProvider` in `packages/ui/src/pages/home/App.tsx` builds a fresh `GatewayProviderConfig` from the dialog draft and then replaces the stored entry outright:

```ts
const provider: GatewayProviderConfig = {
  api_base_url: normalizeProviderBaseUrl(baseUrl),
  api_key: primaryApiKey,
  // ...only the fields the form knows about
};
// ...
config.Providers[providerEditIndex] = provider;
```

`GatewayProviderConfig` also carries `extraBody`, `extraHeaders`, `transformer`, `billing` and the legacy `provider` protocol alias. The backend handles all of them fine — `parseProviders` keeps them, `runtime-topology` forwards them, and the gateway reads `extraBody.default` / `extraBody.byModel[model]` — but the dialog has no field for them, so they are dropped on the next save. `enabled` is already hand-preserved from `existingProvider` on the line above, so the precedent for carrying a field over is there; the others were just missed.

The failure is silent and easy to miss:

```json
{
  "name": "sglang-think",
  "type": "openai_chat_completions",
  "api_base_url": "http://…/v1",
  "models": ["…"],
  "extraBody": { "default": { "reasoning_effort": "high" } }
}
```

Open this provider in the UI, change nothing meaningful — re-type the base URL, tick a model — and save. `extraBody` is gone. No error, no warning, and the UI looks identical because it never displayed the field in the first place. Requests keep succeeding, just without reasoning. The same applies to a `transformer` chain or `extraHeaders` needed by an upstream.

Only the dialog save path is affected. `setProviderEnabled`, the model-metadata editors and the other provider mutations all spread the existing provider and are fine.

## Fix

Carry the unmanaged fields over from the existing provider on edit, via a small pure helper next to the existing `providerCapabilitiesForSave`:

```ts
const provider: GatewayProviderConfig = {
  ...providerManualFieldsForSave(existingProvider),
  api_base_url: normalizeProviderBaseUrl(baseUrl),
  // ...
};
```

The spread comes first so every field the form does manage still wins.

Two deliberate choices:

- The legacy `apiKey` / `apikey` / `baseUrl` / `baseurl` aliases are **not** carried over. The form writes the canonical `api_key` / `api_base_url`, and a surviving alias would shadow the value the user just typed. This is why the fix is an explicit allowlist rather than `{ ...existingProvider, ... }`.
- The helper omits keys the provider never had, so an unrelated edit does not add `undefined`-valued keys to the config draft.

## Tests

Three cases added to `packages/ui/test/integration/providers.test.ts`, alongside the existing `providerCapabilitiesForSave` save-path tests: hand-written fields survive an edit, legacy credential aliases do not, and nothing is invented for a provider that never had these fields.

`npm run test:ui` (183 passing), `npm run test:architecture` and `npm run typecheck` are green.

## Follow-up, not in this PR

This restores the fields but still leaves them uneditable from the UI — `extraBody` and friends remain config-file-only, and a user has no way to see that their provider carries them. An "Advanced" JSON section in the provider dialog would close that gap; happy to open a separate PR if that direction is welcome.

### DIFF
--- a/packages/ui/src/pages/home/App.tsx
+++ b/packages/ui/src/pages/home/App.tsx
@@ -25,7 +25,7 @@ import {
   persistLanguagePreference, PluginInstallCandidate, PluginMarketplaceEntry, PluginRoutingConfigTarget, PluginSettingsDraft, presetCapabilitiesFromDraft,
   probeProviderCandidates, probeProviderDeepLinkPayload, profileAgentLabel, profileAgentOptionsForRuntime, profileDraftWithDetectedAppPath, profileEnvRowsForAgent, ProfileConfig, ProfileOpenSurface, ProfileRuntimeStatus, profileConfigFromDraft, providerAccountApiKeySafetyIssue,
   profileOpenCommandFallback, profileOpenSurfaces, ProviderAccountSnapshot, providerApiKeySafetyIssue, ProviderConnectivityCheckReport, ProviderDeepLinkPayload, ProviderDeepLinkRequest, providerIdentitySafetyIssue, providerProbeCandidates,
-  providerAutoFetchKnownModelsForSave, providerBaseUrl, providerCapabilitiesForProtocols, providerCapabilitiesForSave, providerConnectivityApiKeyFromDraft, providerConnectivityProviderPlugins, providerGlobalBaseUrlForProbe, providerProbeCandidatesApiKeySafetyIssue, providerProbeHasSupportedProtocol, providerProbeInputKey, providerProtocolOptions, providerSelectableProtocolsFromProbe, ProxyNetworkSnapshot,
+  providerAutoFetchKnownModelsForSave, providerBaseUrl, providerCapabilitiesForProtocols, providerCapabilitiesForSave, providerConnectivityApiKeyFromDraft, providerConnectivityProviderPlugins, providerGlobalBaseUrlForProbe, providerManualFieldsForSave, providerProbeCandidatesApiKeySafetyIssue, providerProbeHasSupportedProtocol, providerProbeInputKey, providerProtocolOptions, providerSelectableProtocolsFromProbe, ProxyNetworkSnapshot,
   ProxyStatus, readLanguagePreference, RequestLogListFilter, RequestLogPage, ResolvedLanguage,
   ResolvedTheme, resolvePluginInstallPlan, resolveProviderDeepLinkCatalogModels, removeLocalAgentProviderPluginsForProvider, RouterRule, routingRuleFromDraft, SettingsPageId,
   setProviderPresets, splitLines, translateAppErrorMessage, translateText, TrayBalanceProgressConfig, TrayWidgetConfig,
@@ -1709,6 +1709,7 @@ function App() {
       nextBaseUrl: baseUrl
     });
     const provider: GatewayProviderConfig = {
+      ...providerManualFieldsForSave(existingProvider),
       api_base_url: normalizeProviderBaseUrl(baseUrl),
       api_key: primaryApiKey,
       autoFetchModels: providerDraft.autoFetchModels || undefined,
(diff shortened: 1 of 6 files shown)


songkuan-zheng:
Verified in the real UI, and added an e2e spec (`dc5a971`) so it stays verified.

The spec boots the actual CLI web server against a temp home, seeds a provider carrying `extraBody` / `extraHeaders` / `transformer`, clicks **Providers → Edit → Save** in Chromium, then reads the config back over RPC.

Against the unpatched build:

```
- extraBody:     {"default":{"reasoning_effort":"high"}}
+ extraBody:     undefined
+ extraHeaders:  undefined
+ transformer:   undefined
```

With the fix, all three survive the save. `npx playwright test` is green (6 passed, 4 electron specs skipped in this environment), along with `npm run test:ui` (183), `npm run test:architecture` and `npm run typecheck`.

The CLI web server bootstrap moved out of `cli-web.spec.ts` into `cli-web-runtime.ts` so both specs can start their own instance — no behaviour change to the existing tests.

songkuan-zheng:
Follow-up opened as #1682 — adds Extra request body / Extra request headers JSON boxes to the provider dialog's Advanced settings, so these stop being config-file-only. It is stacked on this branch, so it will shrink to a single commit once this one lands.

musistudio (commented):
Thanks for preserving these hand-written provider fields. The save-path change itself looks reasonable to me: the preserved fields are spread before the canonical UI-managed fields, so `api_key`, `api_base_url`, and `type` still come from the dialog rather than stale legacy aliases.

I found one test issue that should be fixed before relying on the new e2e coverage:

`tests/e2e/provider-edit.spec.ts` does not pin the UI language, but the app defaults to the system language. In a browser/system locale where `navigator.language` starts with `zh`, the app renders `Providers` as `供应商` and the edit button aria-label as `编辑 config-only`. The new test currently hard-codes English selectors:

```ts
await page.getByRole("button", { name: "Providers", exact: true }).click();
const editButton = page.locator(`button[aria-label="Edit ${configOnlyProvider.name}"]:visible`).first();
```

That makes the spec fail on Chinese-locale environments even though the feature works. Please either force the Playwright context locale to English for this test suite, or make these selectors match both English and Chinese like the Save button already does.

I also verified the non-e2e checks locally:

- `git diff --check bb1db63c... dc5a971a...`
- `npm test -w @claude-code-router/ui`
- `npm run typecheck`

The targeted Playwright spec built successfully, but my local machine does not have the required Chromium binary installed, so I could not run the browser assertions end to end.

songkuan-zheng:
Fixed in `1baeccf`: the spec now pins `test.use({ locale: "en-US" })` for the whole suite, so the "Providers" nav and the "Edit <name>" aria-label stay deterministic on every host locale instead of growing dual-language selectors. Verified by running the spec locally against a built CLI: 1 passed (Chromium). `npm run typecheck` and `npm test -w @claude-code-router/ui` (186) remain green.

songkuan-zheng:
Closing as superseded by #1682, which was merged on 2026-08-27 and now contains everything this PR provided — plus a better design for the parts it changed:

- The `providerManualFieldsForSave` mechanism and the en-US locale pin for `tests/e2e/provider-edit.spec.ts` are in main (via #1682, including this branch's test).
- `extraBody`/`extraHeaders` are no longer preserved through the manual-fields list on purpose: the Advanced settings JSON boxes edit them and round-trip them through the draft explicitly (documented in the merged tests, with a reverse assertion that they stay out of the manual fields).
- main's `tests/e2e/provider-edit.spec.ts` is a superset of this branch's (3 tests vs 1).

Thanks @musistudio for merging #1682 — resolving this branch's conflict would have meant reverting the merged design, so closing rather than rebasing. The original bug this PR fixed (saving a provider from the web UI silently dropping `extraBody`/`extraHeaders`) is fixed in main by the Advanced-settings round-trip.

PR #1682, opened 2026-08-17
feat(ui): edit extraBody and extraHeaders from Advanced settings

songkuan-zheng:
> Stacked on #1680 — that PR's two commits are the first two here, because GitHub can only diff against `main`. Review/merge #1680 first and this one shrinks to its last commit.

## Problem

`extraBody` and `extraHeaders` are config-file-only. The gateway reads them (`extraBody.default` / `extraBody.byModel[model]`), `parseProviders` keeps them, and `runtime-topology` forwards them — but the provider dialog renders no field for either, so from the UI they are invisible and uneditable. On a 3.x install where the config lives in `config.sqlite`, the only way to set one is to hand-write the pre-migration `config.json` or write to the database directly.

That is a sharp edge for anything that needs an upstream parameter the dialog does not model. Turning on reasoning for a self-hosted OpenAI-compatible model, for example, is exactly `extraBody: { default: { reasoning_effort: "high" } }` and nothing else — a one-line config change with no UI path to it.

## Change

Two JSON boxes in the **Advanced settings** section, immediately after the usage connector — which already uses this pattern for its request body, so the shape is not new:

```
Advanced settings
  [x] Auto fetch latest models
  [x] Auto detect protocols
  [ ] Fetch usage
  EXTRA REQUEST BODY      OPTIONAL
  ┌──────────────────────────────────────────────┐
  │ {                                            │
  │   "default": { "reasoning_effort": "high" }  │
  │ }                                            │
  └──────────────────────────────────────────────┘
  Merged into every upstream request for this provider.
  Use "default" for all models, or "byModel" to target one.
  EXTRA REQUEST HEADERS   OPTIONAL
  ┌──────────────────────────────────────────────┐
  │ { "x-tenant": "acme" }                       │
  └──────────────────────────────────────────────┘
  Sent with every upstream request for this provider,
  alongside the API key header.
  PROTOCOL DETAILS
  ...
```

- Both open pre-filled with whatever the provider already carries, pretty-printed. An empty box means the field is absent, not `{}`.
- Malformed JSON, or valid JSON that is not an object, blocks the save with an inline error rather than writing something the gateway would ignore.
- `extraBody` / `extraHeaders` leave `providerManualFieldsForSave` (added in #1680) now that the draft round-trips them. `billing`, `provider` and `transformer` stay there — still no UI, still preserved.

## Tests

`packages/ui/test/integration/providers.test.ts` — draft round-trip for both fields, empty-when-unset, and the three rejection cases (invalid JSON, JSON array, JSON string).

`tests/e2e/provider-edit.spec.ts` — drives the real UI in Chromium: opens Advanced settings on a provider carrying `extraBody`, checks the box is pre-filled, edits it, saves, reads the config back over RPC; plus a case asserting a malformed box surfaces the error and leaves the stored value untouched.

`npm run test:ui` (186), `npm run test:architecture`, `npx playwright test` (8 passed, 4 electron specs skipped in this environment) and `npm run typecheck` are green. `npm run test:core` has 5 failures on this branch and the same 5 on an unmodified `main` checkout — pre-existing here, unrelated to these changes.

## Notes

I scoped this to `extraBody` and `extraHeaders` — the two knobs that change what goes out on the wire. `transformer` is the other config-only field a user might want to reach, but it deserves its own treatment rather than a raw JSON box; happy to look at it separately if you want it.


musistudio (commented):
Thanks for adding UI editing for `extraBody` and `extraHeaders`. I think this needs one correction before merge:

`byModel` is not the raw ai-gateway provider config shape.

The new help text says users can use `"default"` for all models or `"byModel"` to target one, and the new test also round-trips this shape:

```json
{
  "byModel": {
    "model-a": {
      "reasoning_effort": "high"
    }
  }
}
```

However, ai-gateway only uses `byModel` after parsing/normalizing the raw config. In the source repo, `src/config.ts` parses raw `extraBody` / `extraHeaders` with `parseModelScopedBody` and `parseModelScopedHeaders`; `hasExplicitModelScope` only treats `default` or actual configured model names as scoped keys. The raw config format documented in `docs/usage.md` is `default` plus model names, not `byModel`.

So the UI should guide users to save this shape instead:

```json
{
  "default": {
    "reasoning_effort": "high"
  },
  "model-a": {
    "reasoning_effort": "max"
  }
}
```

I verified this against the ai-gateway source checkout by calling `parseGatewayConfigFromRaw`: raw `extraBody.default + extraBody["model-a"]` normalizes correctly to internal `byModel["model-a"]`, but raw `extraBody.byModel["model-a"]` normalizes to `byModel.byModel`, so it never applies to `model-a`. For `extraHeaders`, the raw `byModel` object is dropped because header values are expected to be strings/numbers/booleans, not nested objects.

Please update the UI copy, placeholder/examples, and tests to use the raw ai-gateway shape (`default` plus top-level model names), or explicitly translate a UI-level `byModel` object into ai-gateway's raw config shape before saving.

songkuan-zheng:
Thanks — confirmed against the ai-gateway source (parseModelScopedBody scope keys are `default` + model names, and raw `byModel` collapses to `byModel.byModel`). Fixed in `ecec227`: the Advanced settings help text now says to use a model name as the key ("default" for all models), the en/zh i18n rows follow, and the round-trip test seeds the raw shape `{ "default": {...}, "model-a": {...} }` instead of the normalized form. `npm test -w @claude-code-router/ui` is green (186 passed).

PR #1690, opened 2026-08-18
fix(fusion-vision): normalize bare-base64 imageUrl before forwarding

songkuan-zheng:
## What

The virtual-model tool loop hands images to `vision_understand` as a **bare base64 string** in the `imageUrl` argument. `imageInputToUrl` forwarded it verbatim, so strict OpenAI-compatible vision gateways reject the whole request:

- `gateway.theturbo.ai` → HTTP 400 "invalid URL"
- lenient gateways (e.g. maasapi anispark) happen to accept it, which masks the bug

Real-world evidence: a deployment pointed at a strict gateway hit 4× 400 during boot probing and another 400 mid-session (each ~6s wasted before the gateway refused, then a retry).

## Fix

Normalize a non-http(s), non-`data:` `imageUrl` as bare base64 in `imageInputToUrl`, reusing `toDataUrl` — the exact path the `imageBase64` field already takes. Doing it at the MCP boundary covers every source of the argument (loop-emitted and client-issued tool calls) and keeps the tool input schema unchanged.

## Notes

- `data:`-URL input and http(s) input are untouched.
- Bare base64 and its normalized `data:` URI carry identical payload bytes, so any code keying on the post-comma payload is unaffected.

## Testing

- Strict gateway: bare-base64 call previously 400, now 200 (verified against gateway.theturbo.ai).
- Lenient gateway: unchanged.

### DIFF
--- a/packages/core/src/mcp/fusion-vision-mcp.ts
+++ b/packages/core/src/mcp/fusion-vision-mcp.ts
@@ -757,7 +757,15 @@ async function buildImageParts(args: Record<string, unknown>, detail: "auto" | "
 
 async function imageInputToUrl(input: { base64?: string; mimeType?: string; path?: string; url?: string }): Promise<string | undefined> {
   if (input.url) {
-    return input.url;
+    const url = input.url.trim();
+    if (!url.startsWith("data:") && !/^https?:\/\//i.test(url)) {
+      // The virtual-model tool loop emits images as bare base64 in imageUrl.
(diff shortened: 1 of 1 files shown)


songkuan-zheng:
Closing in favor of #1695: its bare-base64 normalization is subsumed as one branch of the input validation in `imageInputToUrl` there (plus repair/drop for truncated base64, local paths, and SVG/XML payloads). The branch `fix/normalize-bare-base64-image-url` stays untouched if the minimal version is ever wanted.

PR #1692, opened 2026-08-18
Drop non-replayable reasoning items from openai_responses input

### LINKED ISSUE #1686: Anthropic thinking content blocks in message history break openai_responses upstream (400) - stripping them makes the same body pass
## Summary

## Environment

- CCR: 3.0.21 (host, macOS)

## Reproduction

|---|---|

```
```

## Root cause

## Suggested fix

dylanpulver:
When a conversation with Anthropic thinking blocks in its history is routed to an openai_responses provider, the upstream answers 400 invalid request and the router silently falls back off the requested model.

The root cause sits in the gateway protocol conversion. Thinking blocks from history are converted into Responses reasoning input items that carry a fabricated rs_ id and plain reasoning text but no encrypted_content, and Responses upstreams reject reasoning input they cannot replay. Reasoning that originally came from a Responses upstream is different: it round-trips through redacted_thinking with its real id and encrypted_content intact, which is why the empirical table in the issue shows redacted_thinking history working while thinking history fails. Stripping the fabricated items was verified in the issue to turn the 400 into a direct 200 on the requested model.

This change fixes it at the CCR provider boundary in packages/core. A new hook registered in createGatewayPlugin runs on the fully translated upstream request and, only when the target provider type is openai_responses and the body is JSON, removes reasoning input items that lack a non-empty encrypted_content string. Reasoning items that do carry encrypted_content pass through untouched, as do other protocols and non-JSON bodies. The boundary is the right place for this repo because the converter itself lives in a dist-only dependency, so this ships now, and if the converter later stops emitting these items the hook becomes a harmless no-op. I can also file the matching guard upstream in the converter if you would rather fix it at the source long term.

Coverage: unit tests exercise dropping thinking-derived reasoning items, preserving encrypted reasoning items, filtering of mixed input arrays, pass-through for non-Responses providers and for non-JSON bodies and for bodies whose input is not an array, and registration of the hook in the gateway plugin. npm run typecheck passes, and so do the core unit and architecture suites.

Fixes #1686


musistudio:
I think this filter is scoped a bit too broadly.

`dropNonReplayableReasoningItems` currently runs for every request targeting an `openai_responses` provider, then removes any `reasoning` input item that does not have non-empty `encrypted_content`. That fixes the Claude Code/Anthropic conversion case, where CCR fabricates non-replayable reasoning items, but it also changes native/pass-through `/v1/responses` behavior.

For a client sending an OpenAI Responses request directly through CCR, reasoning input items are part of the Responses API shape and `encrypted_content` is optional. Dropping those items at the provider boundary can silently remove conversation state that the client intentionally supplied.

Can we narrow this to the Claude Code/Anthropic conversion case, or to the specific fabricated reasoning item shape emitted by the converter? I’d also add a regression test where a native Responses request with a reasoning item lacking `encrypted_content` is left unchanged.

dylanpulver:
Agreed, that was too broad. Rebased and narrowed.

The filter now only runs when the request came in on the Anthropic Messages source adapter (`sourceAdapterKey === "anthropic_messages"` or `sourceProvider === "anthropic"`), matching the gate `applyResponsesToolStrictness` uses. Native `/v1/responses` input items are left alone regardless of `encrypted_content`.

New test: "native Responses requests keep reasoning items without encrypted_content" in `responses-reasoning-filter.test.mjs`, plus cases for Chat Completions conversions and an unknown source protocol. That native test fails if I drop the source gate back out, so it's holding the line.

I went with the source gate over shape-matching because the converter's output is not distinguishable from a client-supplied item: it emits `{ type, id: "rs_" + hex, summary, content: [{ type: "reasoning_text" }] }`, which is the same shape a native client sends, and the fabricated id comes from `randomUUID()` so it looks like a real one. Happy to switch if you'd rather key on the shape.

PR #1695, opened 2026-08-19
fix(fusion-vision): validate and repair image inputs before forwarding

songkuan-zheng:
## What

`imageInputToUrl` in the fusion-vision MCP is the last line before the upstream vision provider. It currently wraps any non-HTTP(S)/non-data value in a blind `data:image/png;base64,...` envelope (#1690) and forwards the rest untouched. Strict upstreams (litellm in front of strict providers) answer malformed `image_url` payloads with a **400 that surfaces to the caller as a raw provider error** — the tool's skip/error messaging never gets a chance.

## Failure shapes seen in production

| Shape | Before | After |
| --- | --- | --- |
| local file path passed as `imageUrl` (also a bare `[media_ref:...]` id) | wrapped as fake base64 → upstream 400 | dropped, value quoted in the skip reason |
| base64 truncated to `len % 4 == 1` (cut mid-image) | forwarded → truncation is unrecoverable, upstream 400 | dropped with reason |
| base64 truncated to `len % 4 == 2/3` | forwarded → 400 or silent corruption | repaired by appending `=` padding (lossless) |
| SVG/XML payload | forwarded → upstream accepts only jpeg/png/gif/webp, 400 | dropped with the supported-format list in the reason |
| mixed `images[]` with one bad entry | whole call poisoned | bad entries skipped, valid ones forwarded |
| all images unusable | raw 400 leak | tool error listing every skip reason |

Data URLs are re-validated the same way (payload is checked/repaired, original media-type label kept). The bare-base64 wrap from #1690 is preserved as one branch of the same check.

## Verification

- 8-case harness against a mock upstream: bare base64 → wrapped `data:image/png;base64,...` (200), path → clean skip (0 upstream calls), `%4==2` → padded (200), `%4==1` → clean skip, SVG → clean skip, `data:` webp → passthrough, mixed array → only valid forwarded, no-image → original error text.
- Live against a real strict chain: stripped-then-padded PNG → 200; SVG → provider's 400 confirmed for the supported-format list quoted in the skip reason (so the reason is accurate, not guessed).

## Notes

- Independent of #1690 (rebased on main): this PR subsumes that one's bare-base64 wrapping as one branch of the same validation, so if both merge this one supersedes it.
- The skip policy for SVG is "drop with reason" rather than convert; if a future provider accepts SVG this branch can trivially forward instead.

- Verification artifacts are in the deployment's patches README; the fix is shipped there since 2026-08-19.

## Review follow-up (2026-08-20)

Maintainer review #1695 caught three defects, fixed in `fc16be5` (+ tests in `f3bff5c`):

1. `imageBase64` with a full `data:...;base64,...` URL was rejected (schema documents both shapes) — data URLs now share the `imageUrl` code path.
2. Local files were forwarded without inspecting contents, so a `.svg` on disk went upstream as a fake raster data URL — file bytes now run through the same content checks.
3. Media-type labels are restricted to jpeg/png/gif/webp: explicit supported label wins, else the sniffed format, else `image/png`.

Verified: 4 new regression tests in the fusion-vision integration suite (7 fusion tests total, 202 core tests, 0 failures) plus `npm run typecheck`.

### DIFF
--- a/packages/core/src/mcp/fusion-vision-mcp.ts
+++ b/packages/core/src/mcp/fusion-vision-mcp.ts
@@ -77,7 +77,7 @@ const visionTool = {
     detail: { enum: ["auto", "low", "high"], type: "string" },
     imageBase64: { description: "Single raw base64 image payload or data URL.", type: "string" },
     imagePath: { description: "Single local image path.", type: "string" },
-    imageUrl: { description: "Single HTTP(S) image URL or data URL.", type: "string" },
+    imageUrl: { description: "Single HTTP(S) image URL, data URL, or bare base64 payload.", type: "string" },
     images: {
       items: objectSchema({
         base64: { type: "string" },
--- a/packages/core/src/mcp/fusion-vision-mcp.ts
+++ b/packages/core/src/mcp/fusion-vision-mcp.ts
@@ -779,6 +779,7 @@ type ImageInputToUrlResult = { url: string } | { skip: string };
  * (irreparably truncated mid-image), and an XML/SVG payload -- which the typical
  * upstream rejects outright (supported formats are jpeg/png/gif/webp), so
  * relabeling it buys nothing. A remainder of 2 or 3 is repairable by padding.
+ * Local files (imagePath/images[].path) go through the same content checks.
  */
 async function imageInputToUrl(input: { base64?: string; mimeType?: string; path?: string; url?: string }): Promise<ImageInputToUrlResult> {
   if (input.url) {
@@ -787,20 +788,7 @@ async function imageInputToUrl(input: { base64?: string; mimeType?: string; path
       return { url };
     }
     if (url.startsWith("data:")) {
-      const comma = url.indexOf(",");
-      if (comma < 1) {
-        return { skip: `malformed data URL (${preview(url)})` };
-      }
-      const header = url.slice(5, comma);
-      if (!/;base64/i.test(header)) {
-        return { skip: `data URL is not base64 (${preview(url)})` };
-      }
-      const mimeType = header.split(";")[0] || undefined;
-      const normalized = normalizeImagePayload(url.slice(comma + 1));
-      if (!normalized) {
-        return { skip: `data URL payload is not usable base64 (${preview(url)})` };
-      }
-      return imageDataUrlOrSkip(normalized, mimeType);
+      return dataUrlResult(url, input.mimeType);
     }
     // Not an HTTP(S) URL and not a data URL. Either a bare base64 payload (the
     // virtual-model tool loop's usual shape; wrapped here because strict gateways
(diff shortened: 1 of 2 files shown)


musistudio:
Thanks for the fix. I think this PR needs a couple of changes before it can merge:

1. `imageBase64` data URLs now get rejected.

   The previous implementation accepted both raw base64 and full `data:image/...;base64,...` strings for `imageBase64`, and the schema still documents that behavior. In the new code, the `input.base64` branch passes the entire value directly into `normalizeImagePayload`, so a valid value like `data:image/png;base64,iVBORw0KGgo=` fails the base64 regex and the tool returns "found no usable image" without calling the upstream provider.

   This is around `packages/core/src/mcp/fusion-vision-mcp.ts:814-819`.

2. Local SVG/XML files still bypass the new validation.

   The `imagePath` branch reads the file and immediately calls `imageDataUrlOrSkip({ payload: buffer.toString("base64"), svg: false }, ...)`. Because it never inspects the decoded file contents, a local `.svg` file is wrapped as `data:image/png;base64,...` and forwarded to the provider. That seems to preserve the exact strict-provider 400 this PR is meant to avoid.

   This is around `packages/core/src/mcp/fusion-vision-mcp.ts:824-828`.

3. MIME labels should probably be restricted to the supported image types.

   The PR description says strict providers support `jpeg`, `png`, `gif`, and `webp`, but the new helper still preserves arbitrary MIME labels from a data URL or `mimeType`. That means inputs like `data:text/plain;base64,<png bytes>` or `mimeType: "image/svg+xml"` with non-SVG bytes can still be forwarded as unsupported media types.

I verified the first two issues with a runtime harness against the compiled PR code: `imageBase64` with a data URL was skipped, and a local SVG file was forwarded upstream as a fake PNG data URL. Existing integration tests and core typecheck still pass, so these cases should be added to the test coverage as regressions.

songkuan-zheng:
Thanks for the careful review — all three points confirmed and fixed (commits `fc16be5` + `f3bff5c`):

1. **`imageBase64` data URLs** — regression confirmed: the validation passed the whole `data:...;base64,...` string to the base64 check. Restored by routing data URLs in `imageBase64` through the same parsed path as `imageUrl` (shared `dataUrlResult()`), so both fields accept the same shapes the schema documents.

2. **Local SVG/XML files** — confirmed: the file bytes were wrapped without inspection, so a `.svg` on disk went upstream as a fake raster data URL. `imagePath`/`images[].path` now run the decoded content through the same checks as every other input (SVG head detection included), so a local SVG is rejected with the same skip reason instead of forwarded.

3. **Media-type labels** — the label is now restricted to the supported raster types: an explicit supported label wins, otherwise the format sniffed from the byte head, otherwise `image/png`. Non-image labels from either the data-URL header or `mimeType` are no longer forwarded (verified: `application/octet-stream` + PNG bytes → `data:image/png;...`, `data:text/plain;base64,<png>` → `data:image/png;...`).

Regression tests added for all three cases plus a positive local-PNG-file case (guard against breaking the path branch); the full core integration suite (202 tests, incl. 7 fusion-vision) and `npm run typecheck` pass locally.

PR #1699, opened 2026-08-19
fix(gateway): pair dispatchers with bundled undici fetch on undici major mismatch (Node >= 25)

krajcik:
# Problem

On Node >= 25 (e.g. Node 26.x, which ships undici 8.x as the built-in fetch) **every upstream provider request fails** with:

```json
{"error":{"message":"All target providers failed.",...,"attempts":[{"stage":"upstream_connect","message":"Failed to reach upstream provider.","status":502}]}}
```

Gateway debug logs (`GATEWAY_LOG_LEVEL=debug`) show the root cause:

```
"details":"fetch failed => invalid onError method (UND_ERR_INVALID_ARG)"
```

This reproduces with any provider (OpenAI, Z.ai, ...) and is independent of proxy settings — the gateway is unusable on current Node.

# Root cause

- Node >= 25 bundles undici 8, while CCR resolves `CCR_UNDICI_MODULE` to its bundled undici 7.x.
- Dispatcher/Handler interfaces are not compatible across undici majors, so a dispatcher created by one undici copy fails inside another copy's fetch with `UND_ERR_INVALID_ARG: invalid onError method`.
- The gateway (ai-gateway) passed its own bundled-undici dispatcher through `globalThis.fetch`, and CCR's generated `gateway-proxy-preload.cjs` injected its own bundled-undici dispatchers into the runtime's built-in fetch — both paths crossed undici copies.

# Fix

The dispatcher/fetch pairing now lives upstream: [The-NeXT-AI/ai-gateway#22](https://github.com/The-NeXT-AI/ai-gateway/pull/22) (released in `@the-next-ai/ai-gateway@1.0.18`) routes dispatcher-carrying requests through the `fetch` of the same undici copy that provides the dispatcher. CCR consumes it:

- Bump `@the-next-ai/ai-gateway` to `^1.0.18` (root, core, cli, lockfile).
- Reduce the generated preload to the minimal CCR-specific proxy/timeout injection:
  - calls that already carry a dispatcher stay on `globalThis.fetch` untouched (their owner is now responsible for pairing — ai-gateway >= 1.0.18 does);
  - calls that receive an **injected** `Agent`/`ProxyAgent` run through the bundled undici module's own `fetch`, so the injected dispatcher and the fetch always come from the same undici copy — no version detection heuristics;
  - calls with no dispatcher to inject keep using the global fetch.
- `undici-proxy-agent.ts` (the packaged shim `CCR_UNDICI_MODULE` can resolve to) now re-exports `fetch` alongside `Agent`/`ProxyAgent`, so the preload never depends on an export the shim does not provide.

# Testing

- Updated `packages/core/test/unit/gateway/gateway-fetch-preload.test.mjs`:
  - injected-dispatcher calls (timeout and proxy routing) go through the bundled undici fetch and never through `globalThis.fetch`;
  - caller-provided dispatchers pass through untouched;
  - no-proxy routing (explicit hosts, loopback, suffix wildcards) unchanged.
- `npm run typecheck` clean; full core test suite has the same three pre-existing `Claude Design` plugin failures as `main`, unrelated to this change.
- Verified end-to-end on Node v26.7.0 (built-in undici 8.9.0, bundled undici 7.29.0): gateway 502s on every request before the patch, all providers answer normally after it.


musistudio:
Thanks for digging into the Node >= 25 / undici major mismatch. I agree with the root cause, but I do not think we should solve this primarily by adding more logic to CCR's generated `globalThis.fetch` monkey patch.

The more robust fix should live upstream in [`The-NeXT-AI/ai-gateway`](https://github.com/The-NeXT-AI/ai-gateway): when `ai-gateway` creates or passes an undici dispatcher, it should call the `fetch` implementation from the same undici package instead of relying on Node's built-in/global `fetch`.

Could you please rework this as follows?

1. Patch `The-NeXT-AI/ai-gateway` so requests that pass `dispatcher` use `undici.fetch` from the same dependency copy that provides `Agent`/`ProxyAgent`/custom dispatchers.
2. Add upstream coverage for Node >= 25 or, at minimum, a test that verifies dispatcher-carrying fetch calls do not go through `globalThis.fetch` when the undici major differs.
3. Release or reference the upstream fix, then update CCR's `@the-next-ai/ai-gateway` dependency.
4. Clean up CCR's preload monkey patch so it only handles the minimal CCR-specific proxy/timeout injection path.

There is also a concrete issue with the current implementation: in packaged CCR, `CCR_UNDICI_MODULE` can resolve to `dist/main/undici-proxy-agent.js`, and that shim currently only exports `Agent` and `ProxyAgent`, not `fetch`. The new fallback path returns `undiciModule.fetch`, so the packaged/runtime path can end up calling `undefined` once the version lookup fails.

Thank you again for your contribution!

krajcik:
The upstream fix is now in The-NeXT-AI/ai-gateway#22: https://github.com/The-NeXT-AI/ai-gateway/pull/22

It pairs dispatcher-carrying requests with that package copy's `undici.fetch` and adds coverage for keeping `globalThis.fetch` out of that path. The full upstream suite passes locally (776 passed, 7 skipped), as do typecheck and build. GitHub Actions is currently awaiting maintainer approval.

Once the upstream change is merged and published, I'll update CCR's `@the-next-ai/ai-gateway` version and reduce this PR's preload back to the minimal CCR-specific proxy/timeout injection. I am intentionally not pinning CCR to a contributor-fork git dependency: the repository does not track `dist`, so that would not produce a usable runtime package.

musistudio:
I have released version 1.0.18 of ai-gateway.

krajcik:
Thank you for the review and for releasing 1.0.18! Pushed 98b30c2 reworking this PR as requested:

- Bumped `@the-next-ai/ai-gateway` to `^1.0.18` (root, core, cli, lockfile) — dispatcher-carrying requests are now paired with their own undici copy upstream, and no longer go through the patched `globalThis.fetch` at all.
- Reduced the preload to the minimal CCR-specific proxy/timeout injection. The only remaining dispatcher handling is for dispatchers the preload itself injects: they are passed to the bundled undici module's own `fetch` (no version-detection heuristics), while caller-provided dispatchers pass through untouched.
- `undici-proxy-agent.ts` now re-exports `fetch`, so the packaged `CCR_UNDICI_MODULE` path can never resolve to a module without it — the `undefined` case you pointed out is gone.

PR description updated to match. Core test suite and typecheck pass (same three pre-existing `Claude Design` plugin failures as `main`).

PR #1702, opened 2026-08-21
fix(gateway): strip Anthropic thinking content blocks for openai upstreams

### LINKED ISSUE #1686: Anthropic thinking content blocks in message history break openai_responses upstream (400) - stripping them makes the same body pass

Burlesque1:
## Summary

OpenAI chat/responses upstreams reject Anthropic `thinking` and `redacted_thinking` content blocks in message history with HTTP 400. Previously `stripUnsupportedOpenAiRequestParameters` only removed top-level `thinking`/`reasoning_split` fields; it now also strips `type: "thinking"` / `type: "redacted_thinking"` blocks from each message's `content` array before forwarding to OpenAI targets.

Fixes #1686

## Changes

- `packages/core/src/gateway/upstream/executor.ts` — extend `stripUnsupportedOpenAiRequestParameters` to also strip thinking content blocks from `messages[].content[]`; export the function for unit testing. Non-array (string) content is left untouched.
- `packages/core/test/unit/gateway/upstream/executor.test.mjs` — 4 unit tests covering thinking/redacted_thinking stripping, the top-level `thinking`/`reasoning_split` regression, no-op passthrough when there is nothing to strip, and OpenAI-style string content.

## Testing

- 4/4 new unit tests pass.
- Full core suite: no regressions introduced (599 tests; the only 3 failures are pre-existing claude-design plugin path issues in the local environment, unrelated to this change).

### DIFF
--- a/packages/core/src/gateway/upstream/executor.ts
+++ b/packages/core/src/gateway/upstream/executor.ts
@@ -789,15 +789,56 @@ function usageAwareOpenAiChatAttemptBody(input: {
 }
 
 
-function stripUnsupportedOpenAiRequestParameters(body: Buffer | undefined): Buffer | undefined {
+export function stripUnsupportedOpenAiRequestParameters(body: Buffer | undefined): Buffer | undefined {
   const parsedBody = parseJsonObjectSafe(body);
-  if (!parsedBody || (!("thinking" in parsedBody) && !("reasoning_split" in parsedBody))) {
+  if (!parsedBody) {
     return body;
   }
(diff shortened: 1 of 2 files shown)


Burlesque1:
Superseded by #1784, which includes this thinking-block strip plus the Responses reasoning-content strip in the same function.

PR #1703, opened 2026-08-21
fix(gateway): translate request body across cross-protocol fallback

### LINKED ISSUE #1615: Bug: Cross-protocol fallback skips body re-translation causing HTTP 400
### Bug Description
When a fallback chain contains models using different protocols (e.g., anthropic_messages followed by openai_responses), retrying after a retryable error (429/5xx) from the first model returns HTTP 400 on the next attempt because the fallback executor does not re-run protocol translation.

### Steps to Reproduce
1. Configure a fallback chain with alternating protocols, e.g.:
   - Model A: anthropic_messages (e.g., bytedance7::anthropic_messages/ep-...)
   - Model B: openai_responses (e.g., bytedance2::openai_responses/ep-...)
   - Model C: anthropic_messages
2. Send a request with Anthropic-native fields (e.g., `thinking: {type: "adaptive"}`, `context_management`, `output_config`, tool definitions, `system` as array with `cache_control`)
3. Trigger Model A to return 429 (e.g., via rate limiting by sending concurrent requests)

### Expected Behavior
Fallback to Model B should work: the request body should be re-translated from Anthropic Messages format to OpenAI Responses format before sending.

### Actual Behavior
- Attempt 1 (Model A, anthropic_messages): 429 rate limited
- Attempt 2 (Model B, openai_responses): **400 Bad Request**
- Attempt 3 (Model C, anthropic_messages): 200 OK

Model B receives an Anthropic-format body (with `thinking`, `system` as array, etc.) directly sent to the OpenAI Responses endpoint, which rejects unknown fields with HTTP 400.

### Root Cause
In the fallback executor, when retrying with a different provider/model, only the `model` field in the body and routing target are replaced. The `capability-routing` translation step (seq=5 in the route trace) is NOT re-executed, so the request body remains in the original protocol format and is incompatible with the fallback model's protocol.

### Evidence
Route trace from a real failed request (request_id 86984338-e938-4f62-8f10-537f9390eff4):
```
seq=5: provider.capability-routing → translates body to anthropic for bytedance7::anthropic_messages/7pszs
seq=8: attempt #1 → 429
seq=9: attempt #2 prepare → only replaces /body/model with bytedance2::openai_responses/hn6s9 (no re-translation!)
seq=10: attempt #2 → 400
seq=11: attempt #3 prepare → replaces model back to bytedance7::anthropic_messages/4hrpz
seq=12: attempt #3 → 200
```

Direct requests to Model B (bypassing fallback, going through full translation) work correctly for all tested scenarios: simple requests, tools, thinking=adaptive, metadata, system with cache_control, 137 tools, 100k+ token payloads, streaming.

### Workaround
Reorder the fallback chain to group models by protocol (all anthropic_messages first, then all openai_responses/chat_completions) so cross-protocol fallback does not happen until all same-protocol models are exhausted.

### Proposed Fix
When the fallback target uses a different protocol/capability than the previous attempt, the executor must re-run the full body translation pipeline (`provider.capability-routing` hop) to convert the body to the new protocol format before sending.

Burlesque1:
## Summary

When a request arrives on one protocol (e.g. `/v1/messages`, Anthropic) but the fallback target resolves to a provider speaking a different protocol (e.g. `openai_chat_completions`), the body was forwarded unchanged with only the `model` field rewritten, causing HTTP 400 upstream.

This PR adds a defensive Anthropic <-> OpenAI chat request-body translation, applied per fallback attempt before forwarding, keyed by the resolved target provider protocol.

Fixes #1615

## Changes

- `packages/core/src/gateway/upstream/body-translate.ts` (new) — `translateBodyForProtocol` plus the Anthropic <-> OpenAI chat translators. Covers the system prompt, text content, tool_use/tool_result, tool definitions, stop/stop_sequences, and sampling params. Unsupported protocol pairs and any translation failure return the original body, so a fallback attempt degrades to today's behavior instead of breaking.
- `packages/core/src/gateway/upstream/executor.ts` — in `prepareUpstreamCredentialAttempt`, translate the body from the client protocol to the resolved target provider protocol before OpenAI-specific sanitization.
- `packages/core/test/unit/gateway/upstream/body-translate.test.mjs` — 6 unit tests (both translation directions, no-op on matching/unsupported/missing protocols, and preservation of incidental fields such as `stream_options`).

## Testing

- 6/6 new unit tests pass.
- `router-builtins.test.js`: 78/78 pass — no regression.
- Full core suite: no regressions introduced (the only failures are pre-existing claude-design plugin path issues in the local environment, unrelated to this change).

### DIFF
--- a/packages/core/src/gateway/upstream/body-translate.ts
+++ b/packages/core/src/gateway/upstream/body-translate.ts
@@ -1,0 +1,398 @@
+import type { GatewayProviderProtocol } from "@ccr/core/contracts/app";
+import { parseJsonObjectSafe, serializeJsonBody } from "@ccr/core/gateway/http/body";
+import { isRecord, stringValue } from "@ccr/core/gateway/internal/value";
+
+/**
+ * Cross-protocol fallback body translation (issue #1615).
+ *
+ * When a request arrives on one protocol (e.g. `/v1/messages`, Anthropic) but the
+ * fallback target resolves to a provider speaking a different protocol (e.g.
+ * `openai_chat_completions`), the body is currently forwarded unchanged with only the
+ * `model` field rewritten — the upstream rejects it with HTTP 400. These helpers translate
+ * the request body between the Anthropic messages format and the OpenAI chat completions
+ * format so a cross-protocol fallback actually works.
+ *
+ * Only the protocol pair that covers the common fallback scenarios is implemented
+ * (`anthropic_messages` <-> `openai_chat_completions`). Other pairs (openai_responses,
+ * gemini, ...) fall through unchanged. Translation is defensive: any structure it cannot
+ * map is carried over or dropped conservatively, and it never throws — on failure the
+ * original body is returned so the fallback degrades to today's behavior instead of crashing.
+ */
+
+export function translateBodyForProtocol(
+  body: Buffer | undefined,
+  clientProtocol: GatewayProviderProtocol | undefined,
+  targetProtocol: GatewayProviderProtocol | undefined
+): Buffer | undefined {
+  if (!body || !clientProtocol || !targetProtocol || clientProtocol === targetProtocol) {
+    return body;
+  }
+  try {
+    const parsedBody = parseJsonObjectSafe(body);
+    if (!parsedBody) {
+      return body;
+    }
+    let translated: Record<string, unknown> | undefined;
+    if (clientProtocol === "anthropic_messages" && targetProtocol === "openai_chat_completions") {
+      translated = anthropicToOpenAiChat(parsedBody);
+    } else if (clientProtocol === "openai_chat_completions" && targetProtocol === "anthropic_messages") {
+      translated = openAiChatToAnthropic(parsedBody);
+    }
+    return translated ? serializeJsonBody(translated) : body;
+  } catch {
+    // Never let a translation failure break a fallback attempt.
+    return body;
+  }
+}
+
+function anthropicToOpenAiChat(body: Record<string, unknown>): Record<string, unknown> | undefined {
+  // Copy everything through and only replace the fields whose shape differs between
+  // protocols. This preserves incidental fields (stream, stream_options, temperature,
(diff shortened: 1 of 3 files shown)


musistudio:
Thanks for working on this and for adding regression tests. The issue is real, and this PR helped make the cross-protocol fallback problem much clearer.

But I think the protocol-body conversion should be fixed in [`The-NeXT-AI/ai-gateway`](https://github.com/The-NeXT-AI/ai-gateway) rather than implemented inside CCR.

The reason is that CCR still forwards the request to the core gateway using the original client route, for example `/v1/messages`. That route selects the `anthropic_messages` source adapter in `ai-gateway`, and `ai-gateway` already owns the `source protocol -> StandardRequest -> target protocol` conversion flow. If CCR rewrites the body before forwarding, we can end up sending an OpenAI-shaped body through the Anthropic source adapter, which risks double conversion or incorrect parsing.

I think the better fix is to keep CCR responsible for fallback/provider/model/header selection, and let `ai-gateway` handle protocol conversion for each target attempt. If reusable conversion is needed outside the server path, `ai-gateway` should expose a shared conversion API based on its existing adapters instead of duplicating protocol mappers here.

So I do not think we should merge this implementation as-is, but I really appreciate the contribution and the investigation. It would be great to move the fix upstream so CCR and any other consumer can benefit from the same conversion logic.

Burlesque1:
Thanks for the careful review — you're right. CCR forwards to the core gateway at the original client route (e.g. `/v1/messages`), and ai-gateway already owns the inbound-protocol → StandardRequest → outbound-protocol conversion via its adapters (handler.ts `toStandardRequest` → `buildRequestFromStandard`). Pre-translating the body in CCR would feed an OpenAI-shaped body through the anthropic_messages inbound adapter, risking double conversion / incorrect parsing.

I verified the claim against the code and agree: protocol conversion belongs in ai-gateway, not CCR. Withdrawing this implementation.

Issue #1628, opened 2026-08-03
claude-code-oauth provider uses a static embedded access token for upstream auth — expires (~8h) and 401s; dynamic resolution exists but is never used


ezra-quemuel:
# claude-code-oauth provider applies a static embedded access token for upstream auth — it expires (~8h) and 401s; dynamic resolution exists but is never used

## Environment

- CCR version: **3.0.18** (npm CLI `@musistudio/claude-code-router`, headless Linux, `ccr serve` under systemd)
- Node: v22.23.2

## Summary

The Claude Code OAuth provider (imported via **Import Local Agent Login**) authenticates upstream requests to `api.anthropic.com` with a **static access token embedded in the provider plugin's `auth.headers`**. That token expires (~8h lifetime) and is also invalidated whenever the interactive Claude Code CLI refreshes the shared token family in `~/.claude/.credentials.json`. Result: the provider starts 401ing until the plugin config is manually updated, and every request routed to it fails or falls back.

The Codex provider does **not** have this problem: its `codexOauth` plugin carries a `refreshToken` and refreshes dynamically (`refreshIfMissingAccessToken`). There is no equivalent dynamic path for `claude-code-oauth`.

## Evidence

**1. The imported plugin is static-only.** `importLocalAgentProvider` for `claude-code-api` returns plugins shaped:

```json
{
  "auth": {
    "headers": {
      "authorization": "Bearer <accessToken at import time>",
      "anthropic-beta": "oauth-2025-04-20"
    },
    "removeHeaders": ["x-api-key"],
    "strict": true
  },
  "key": "ccr-local-agent-...-claude-code-oauth",
  "providerName": "..."
}
```

No `sourceFile`, no `refreshToken`, no expiry — just the frozen header.

**2. The gateway's upstream auth flow never invokes the dynamic resolver that exists in the codebase.** Instrumentation (logging injected into the running bundle):

- `c9e` (`async function c9e(e,t){for(let r of e.providerPlugins??[])… if(n.includes("claude-code-oauth"))return b9e(r)`) — **never fires** for upstream model requests.
- `b9e` (`return{apiKey:IA()?.accessToken||…}` — `IA()` re-reads `~/.claude/.credentials.json` live) — **never fires**.
- The `@the-next-ai/ai-gateway` bundle (the spawned upstream dispatcher) contains **zero** occurrences of `claude-code-oauth`; it applies the plugin's static `auth.headers` as-is.

[...]

PR #1704, opened 2026-08-21
test: cover Claude Code path in local agent auth provider hook

kskadart:
The local agent auth hook re-reads the on-disk Claude Code token on every request, so a token rotation is picked up without a gateway restart. That behaviour arrived in 561c2d8 and is what resolved the problem reported in #1628.

The test file added alongside it only exercises the Grok path, so the Claude Code branch is currently untested. A future refactor could regress it silently.

This adds a single test mirroring the existing Grok case: write credentials, authenticate, rotate the token on disk without restarting, authenticate again, and assert the rotated token is returned along with the anthropic-beta header.

No production code is touched.

Verification:
- npm run typecheck: clean
- npm run test:packages: cli 5/5, core 787/791, electron 28/28, ui 181/181. The 4 core failures are pre-existing and identical on a clean main checkout.


kskadart:
Superseded by #1705 -- rebranched to match the repository's fix/ branch naming convention.

PR #1711, opened 2026-08-23
fix: invalidate Claude gateway model discovery cache when the profile allowlist changes

diogomcd:
## Problem

When a model is removed from a profile's **Allowed model list** in CCR, the removed model keeps appearing in the Claude Code CLI and the Claude App model picker. Adding models works, but removing them does not.

The allowlist itself is applied correctly: the CCR gateway returns the correct `/v1/models` and `/api/claude_cli/bootstrap` responses. The stale models come from a **cache owned by the Claude client**, not from CCR.

## Root cause

With `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1`, Claude Code fetches the gateway model list and persists it on disk:

- CLI: `~/.claude/cache/gateway-models.json`
- App: the Electron/Chromium cache inside the profile user-data dir

That cache is keyed by the gateway base URL. Editing the allowlist does **not** change the base URL, so the client never re-fetches and keeps showing the previous (stale) snapshot. CCR's profile apply flow wrote `settings.json`, the wrapper, and the auto-compact cache, but never invalidated this model-discovery cache.

## Solution

Invalidate the client's model-discovery cache only when the effective discoverable model list changes for a profile — not on every apply.

- Compute a stable fingerprint of the profile's effective discoverable models using `filterModelIdsForProfile(config, availableGatewayModelIds(config), profile)`. This captures allowlist changes as well as provider/virtual-model changes.
- Persist the fingerprint per profile in `<config-dir>/claude-model-discovery-fingerprint.json`.
- On profile apply, when the fingerprint changes: remove the CLI cache `~/.claude/cache/gateway-models.json` (derived from the profile's `settings.json` directory), and refresh the Claude App model-discovery cache via `refreshClaudeAppModelDiscoveryCache(...)` for the profile's user-data dir.

### Files changed

- `packages/core/src/profiles/service.ts` — fingerprint + conditional cache invalidation.
- `packages/core/src/agents/claude-app/gateway-service.ts` — export the existing `refreshClaudeAppModelDiscoveryCache` helper.
- `packages/core/test/integration/profiles/profile-service.test.mjs` — regression test.

## Test plan

- `npm run typecheck`
- `npm run test:architecture`
- `npm test -w @claude-code-router/core` (profile-service test)

New regression test verifies that the cache is: removed on first apply, preserved when the allowlist is unchanged, and removed again when the allowlist changes.

## Notes

- The first apply after this change is treated as a change (no prior fingerprint), so it removes the cache once. Subsequent applies only touch the cache when the discoverable model list actually changes.
- This covers both the Claude Code CLI and the Claude App surfaces.

### DIFF
--- a/packages/core/src/agents/claude-app/gateway-service.ts
+++ b/packages/core/src/agents/claude-app/gateway-service.ts
@@ -172,7 +172,7 @@ function claudeAppGatewayModelsVersion(endpoint: string, models: ClaudeAppGatewa
     .slice(0, 16);
 }
 
-function refreshClaudeAppModelDiscoveryCache(dataDir: string): void {
+export function refreshClaudeAppModelDiscoveryCache(dataDir: string): void {
   for (const relativePath of [
     "Cache",
     "Code Cache",
--- a/packages/core/src/gateway/features/model-discovery.ts
+++ b/packages/core/src/gateway/features/model-discovery.ts
@@ -111,7 +111,11 @@ export function createGatewayModelsResponse(config: AppConfig, headers: Incoming
 
 
 export function createClaudeCliBootstrapResponse(config: AppConfig, apiKey?: ApiKeyConfig): Record<string, unknown> {
-  const profile = profileForApiKey(config, apiKey);
+  return createClaudeCliBootstrapPayload(config, profileForApiKey(config, apiKey));
+}
+
+
+function createClaudeCliBootstrapPayload(config: AppConfig, profile?: ProfileConfig): Record<string, unknown> {
   const windows = createClaudeCliAutoCompactWindows(config, profile);
   return {
     additional_model_options: createClaudeCliAdditionalModelOptions(config, profile),
@@ -119,6 +123,19 @@ export function createClaudeCliBootstrapResponse(config: AppConfig, apiKey?: Api
     client_data: {
       rowan_thicket: { ...windows }
     }
+  };
+}
+
+
+export function claudeClientDiscoveryPayloads(
+  config: AppConfig,
+  options: { contextArchiveCompact?: boolean; profile?: ProfileConfig } = {}
+): Record<string, unknown> {
+  const { contextArchiveCompact, profile } = options;
+  return {
+    bootstrap: createClaudeCliBootstrapPayload(config, profile),
+    claudeApp: createClaudeAppGatewayModelsResponse(config, { contextArchiveCompact, profile }),
+    claudeCode: createClaudeAppGatewayModelsResponse(config, { claudeCode: true, contextArchiveCompact, profile })
   };
 }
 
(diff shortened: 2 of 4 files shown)


musistudio:
Thanks for the fix. I found two issues worth addressing before relying on this cache invalidation behavior:

1. `claudeProfileModelDiscoveryFingerprint()` only hashes the filtered model IDs, but the client cache stores the full Claude discovery/bootstrap payload. Metadata-only changes such as `modelMetadata.contextWindow`, `maxOutputTokens`, reasoning levels, image capability, or display-name/capability changes can alter `/v1/models` and `/api/claude_cli/bootstrap` without changing the ID list, so the stale client cache will survive. I verified a scratch case where the filtered IDs stayed `['Provider/alpha']` while the discovery response changed `max_input_tokens`. Consider hashing the actual generated discovery/bootstrap payload for the profile, or at least the model metadata inputs that shape it.

2. The new invalidation block runs before the existing `try` that turns profile-apply failures into an `ok: false` status. If removing the cache or writing `claude-model-discovery-fingerprint.json` fails because of permissions or an unusual filesystem state, `applyProfileConfig()` can reject and stop later profiles instead of reporting a contained Claude Code profile error. It would be safer to move this block inside the existing error boundary or wrap it with the same status-return behavior.

diogomcd:
**1. Fingerprint now covers the full discovery payload (6040d66)**

`claudeProfileModelDiscoveryFingerprint()` no longer hashes just the filtered model IDs. It now generates the actual client-facing payloads through a new `claudeClientDiscoveryPayloads()` helper in `model-discovery.ts`: the `/api/claude_cli/bootstrap` response plus both `/v1/models` variants (Claude App and Claude Code user agent), including the context-archive compact flag, and hashes their JSON serialization with SHA-256. Since these are exactly the responses the clients cache, any change that alters them triggers invalidation, even when the ID list stays the same: `modelMetadata` fields like `max_input_tokens`/`context_window`, reasoning levels, capabilities, display names, auto-compact windows, and context archive toggles.

To make this possible without touching the request paths, `createClaudeCliBootstrapResponse()` was split into a thin wrapper over a payload builder, and `refreshClaudeAppModelDiscoveryCache()` was exported from `gateway-service.ts`.

**2. Invalidation moved inside the existing error boundary (23592a8)**

The block now sits at the top of the existing `try` in `applyClaudeCodeProfile()`. If deleting the gateway cache file or writing `claude-model-discovery-fingerprint.json` throws (permissions, unusual filesystem state), the profile returns `ok: false` with the formatted error instead of letting `applyProfileConfig()` reject, so profiles after it still get applied.

Two integration tests cover the behavior:
- one asserts the cache survives an apply with an unchanged payload, is invalidated by a metadata-only change (`modelMetadata.contextWindow`) with an untouched allowlist, and by an allowlist change;
- the other replaces the fingerprint state file with a directory to force a write failure and asserts the claude-code profile reports `ok: false` while the following codex profile still applies.

PR #1718, opened 2026-08-25
feat: allow explicit context window overrides for local Codex providers

diogomcd:
## Motivation

On August 16, 2026, OpenAI started exposing a 1M token context window for Codex models through the ChatGPT backend. CCR still clamps every imported Codex GPT-5 context window to hardcoded values (`368_000` for gpt-5.6, `256_000` for other GPT-5 models), and these caps are reapplied at runtime on top of whatever the user saves.

The practical result before this PR: typing `1000000` into the context window field of a local Codex provider looks accepted, then silently reverts to `368000`. Downstream clients follow along: the Codex CLI catalog reports the capped value and Claude Code model discovery reports a reduced window. The cap made sense while backend-reported windows were unreliable, but with the real 1M window now available it blocks legitimate configurations.

## The feature

A new optional `contextWindowPinned?: boolean` flag on `ProviderModelMetadata`. When the user edits the context window field in the provider form, the value is recorded as pinned. Every code path that previously overwrote imported context metadata now skips or restores pinned values. Models without the flag behave exactly as before.

## Why one guard was not enough

Investigating the report surfaced three independent layers that could discard or reduce a user-entered window, all fixed here:

1. **Imported cap overrides.** `codexImportedModelMetadataWithContextOverrides` (provider normalization and local catalog import) and `codexProviderModelMetadataFor` (Codex CLI catalog build) rewrote `contextWindow`/`maxContextWindow` with the fixed caps. Both now skip models whose metadata carries the pin.

2. **Metadata merges.** The auto-refresh merge (`mergeProviderModelMetadata`) spread discovered metadata over current metadata, so the imported 368k overwrote a saved 1M on every background refresh, dropping the flag itself. The UI probe merge (`mergeModelMetadata`) replaced whole metadata objects, which is why the form showed 1000000 on open and flipped to 368000 moments later. Both merges now restore the pinned window and flag after merging.

3. **Effective context window percent.** Even with the value preserved, discovery multiplied it by `effectiveContextWindowPercent` imported from the backend (95 by default), so a pinned 1M reached Claude Code as 950k. When pinned, both the model discovery path and the Codex catalog item report the window in full (percent treated as 100).

## Behavior

| Scenario | Result |
| --- | --- |
| No pin set | Unchanged: caps still clamp imported GPT-5 metadata |
| Pin set | Exact user value in provider config, Codex CLI catalog, and Claude Code discovery |
| "Use preset" in the form | Clears value and flag together |

## Testing

- New unit test: `codex catalog reports pinned context windows in full` covers the pin-plus-percent case (pinned 1M with a 95% imported percent publishes 1000000 at 100%).
- New unit test: `codex catalog keeps pinned context windows above imported caps` and its provider-level counterpart cover cap skipping during normalization and catalog build.
- All pre-existing cap tests pass unchanged, confirming behavior for unpinned models is intact.
- `npm run typecheck` clean; core agent/provider/discovery suites and the UI provider integration suite pass.

### DIFF
--- a/packages/core/src/agents/claude-app/gateway-routes.ts
+++ b/packages/core/src/agents/claude-app/gateway-routes.ts
@@ -1,5 +1,5 @@
 import type { AppConfig } from "@ccr/core/contracts/app";
-import { availableGatewayModelIds, normalizeProfileScopeValue } from "@ccr/core/contracts/app";
+import { availableGatewayModelIds, effectiveContextWindowPercentFor, normalizeProfileScopeValue } from "@ccr/core/contracts/app";
 import { findModelCatalogEntry, findProviderModelCatalogEntry, type ModelCatalogEntry } from "@ccr/core/gateway/model-catalog";
 import { modelRegistryForConfig } from "@ccr/core/routing/model-registry";
 import { resolveUsageModelAttribution } from "@ccr/core/usage/model-attribution";
@@ -215,7 +215,7 @@ function claudeAppGatewayProviderSupportsOneMillionContext(
   if (!contextWindow) {
     return undefined;
   }
-  const effectivePercent = percentage(metadata?.effectiveContextWindowPercent) ?? 100;
+  const effectivePercent = effectiveContextWindowPercentFor(metadata) ?? 100;
   return Math.floor((contextWindow * effectivePercent) / 100) >= 1_000_000;
 }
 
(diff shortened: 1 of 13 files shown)


musistudio:
I think we should remove the hardcoded imported Codex context-window cap.

diogomcd:
I read that as drop the cap entirely rather than add a way around it, so that's what I did here.

The hardcoded imported context window cap for Codex GPT-5 models is gone: the 368_000 / 256_000 constants, codexImportedModelContextWindow, the metadata override applied during provider normalization and local catalog import, and the rewrite in the Codex CLI catalog build. Imported metadata now flows through untouched, so whatever the backend reports is what CCR uses. Existing tests that asserted the capped values were rewritten to assert the reported values instead.

PR #1723, opened 2026-08-25
fix(gateway): opt OpenAI Responses tools out of implicit strict mode

### LINKED ISSUE #1722: [Bug] OpenAI Responses models cannot call `Monitor`: both `command` and `ws` are emitted
# Summary

After `Monitor` is made discoverable in a CCR-routed Claude Code session, an OpenAI Responses model repeatedly produces invalid `Monitor` arguments containing both mutually exclusive sources:

```json
{
  "description": "changes to file.txt",
  "timeout_ms": 300000,
  "persistent": true,
  "command": "inotifywait ...",
  "ws": {
    "url": "ws://placeholder",
    "protocols": []
  }
}
```

Claude Code rejects every call with:

```text
InputValidationError: exactly one of command or ws
```

The tool never starts. Using Claude Code's Bash tool with `run_in_background` works in the same session, isolating the problem to tool-schema/argument translation rather than local process execution.

# Environment

- CCR: 3.0.22
- Claude Code: 2.1.245
- Provider type: `openai_responses`
- Provider model: `gpt-5.6-sol`
- Upstream adapter package bundled by CCR: `@the-next-ai/ai-gateway` 1.0.18
- OS: Linux
- `ENABLE_TOOL_SEARCH=1`

# Steps to reproduce

1. Launch Claude Code through a CCR profile using an OpenAI Responses model.
2. Ensure `Monitor` is discoverable, then ask:

   ```text
   Create file.txt with "hello" and monitor it for changes.
   ```

3. Claude loads `Monitor` through `ToolSearch`.
4. Observe the repeated validation error.

# Expected behavior

For a command-backed monitor, the model call should contain `command` and omit `ws` entirely. For a WebSocket-backed monitor, it should contain `ws` and omit `command` entirely.

# Actual behavior

In one captured session the model attempted `Monitor` seven times. Every attempt contained both keys. It changed the WebSocket placeholder between attempts (`""`, localhost, example.com, and similar values) but never omitted `ws`, even after receiving the exact runtime validation error.

The same session also emitted an empty optional `pages` property on a normal `Read` call, suggesting a broader optional-parameter normalization problem on the OpenAI Responses path.

# Root-cause evidence

The Monitor schema sent by Claude Code declares:

- required: `description`, `timeout_ms`, `persistent`
- optional: `command`, `ws`
- prose: `ws` "Cannot be combined with command"

It does not encode Claude Code's runtime requirement that exactly one source must be supplied.

CCR's bundled OpenAI Responses target adapter passes the source `input_schema` through as `parameters` and only includes `strict` when the source tool explicitly supplies it. Claude Code's Anthropic-style tool schema does not supply `strict`.

OpenAI Responses attempts to normalize tool schemas into strict mode when `strict` is omitted. Official documentation says that if omitted, Responses will try to make the schema strict-compatible and only fall back to non-strict mode when that is impossible; to opt out and keep non-strict, best-effort function calling, callers must set `strict: false` explicitly. Strict function schemas require every property to be required. This is consistent with the model populating every optional property.

Reference: https://developers.openai.com/api/docs/guides/function-calling#strict-mode

# Suggested resolution

## Minimum general fix

When translating a source tool that does not explicitly request strict mode, emit:

```json
{
  "strict": false
}
```

instead of omitting `strict`. This preserves optional Anthropic tool parameters rather than allowing the Responses API to normalize them implicitly.

The relevant mapping currently lives in the bundled `@the-next-ai/ai-gateway` OpenAI Responses target adapter, so CCR may need to coordinate an upstream dependency fix and then upgrade the bundled version.

## Robust Monitor-specific fix

Do not rely only on prose for the `command`/`ws` exclusive choice. Translate Monitor into an OpenAI-compatible discriminated source union, or two unambiguous target tools, and map the selected representation back to Claude Code's original `Monitor` call.

A nested discriminated `anyOf` is compatible with OpenAI Structured Outputs as long as the root remains an object and each branch follows the supported schema subset:

https://developers.openai.com/api/docs/guides/structured-outputs

A response-side guard may also reject or safely normalize malformed calls, but it should not silently choose a branch if both contain meaningful user-directed values.

# Acceptance tests

1. An Anthropic-style tool with optional fields is translated with explicit, intentional strictness.
2. After deferred discovery, command-backed Monitor calls omit `ws`.
3. WebSocket-backed Monitor calls omit `command`.
4. Claude Code accepts the translated call without `InputValidationError`.
5. Changing a watched file produces a Monitor notification through CCR.
6. `TaskStop` cancels the running monitor.

# Related issues

- CCR #343 is a generic older report about invalid tool parameters, but does not cover Monitor or the OpenAI Responses strictness behavior.
- #1721 covers why Monitor is missing in isolated CCR profiles before this argument-translation bug can be reached.

gtapps:
## Summary

OpenAI Responses treats omitted `strict` as an attempt to make the tool schema strict. Claude Code tools do not set `strict`, so optional fields (Monitor `command`/`ws`, Read `pages`) get forced on and Claude Code rejects Monitor with `exactly one of command or ws`.

This stamps `strict: false` on outbound Responses function tools that do not already have a boolean `strict`, using the existing CCR provider-hook path.

Closes #1722

## Changes

- Add `applyResponsesToolStrictness` and register it as `ccr-responses-tool-strictness` next to session affinity.
- Leave explicit `strict: true`/`false`, non-function tools, Chat Completions, and non-JSON bodies unchanged.
- Recurse into `type: namespace` nested function tools.

## Test plan

- `node --test .test-dist/core/test/gateway/responses-tool-strictness.test.js`: 8 passed, 0 failed.
- `npm run typecheck`: exit 0.
- `npm run test:core`: new tests passed. Three bundled-plugin tests failed on this machine; the same three fail on `main` without this change (`known bundled plugins without persisted permissions...`, Claude Design split-plugin migration, Claude Design runtime plugin config).
- No live Claude Code Codex session (usage-limited; not in this PR's scope).

### DIFF
--- a/packages/core/src/gateway/core-runtime/responses-tool-strictness.ts
+++ b/packages/core/src/gateway/core-runtime/responses-tool-strictness.ts
@@ -9,6 +9,8 @@ type UpstreamRequest = {
 };
 
 export type ResponsesToolStrictnessInput = {
+  sourceAdapterKey?: string;
+  sourceProvider?: string;
   targetProviderConfig?: {
     type?: string;
   };
@@ -18,12 +20,19 @@ export type ResponsesToolStrictnessInput = {
 /**
  * OpenAI Responses treats omitted `strict` as "try to make this schema strict".
  * Claude Code tools leave `strict` unset, so optional fields get forced on.
+ * Only converted Anthropic/Claude Code requests are stamped. Native
+ * `/v1/responses` pass-through keeps omitted `strict` as API semantics.
  * Explicit boolean strictness is left alone. Other protocols pass through.
  */
 export function applyResponsesToolStrictness(input: ResponsesToolStrictnessInput): UpstreamRequest {
   const upstreamRequest = input.upstreamRequest;
   const providerType = input.targetProviderConfig?.type?.trim().toLowerCase();
   if (providerType !== "openai_responses") {
+    return upstreamRequest;
+  }
+  const sourceKey = input.sourceAdapterKey?.trim().toLowerCase();
+  const sourceProvider = input.sourceProvider?.trim().toLowerCase();
+  if (sourceKey !== "anthropic_messages" && sourceProvider !== "anthropic") {
     return upstreamRequest;
   }
   const body = upstreamRequest.body;
--- a/packages/core/src/gateway/core-runtime/responses-tool-strictness.ts
+++ b/packages/core/src/gateway/core-runtime/responses-tool-strictness.ts
@@ -1,0 +1,80 @@
+import { isRecord } from "@ccr/core/gateway/internal/value";
+
+type UpstreamRequest = {
+  body?: unknown;
+  bodyEncoding?: "bytes" | "form" | "json" | "none" | "text";
+  headers?: Record<string, string>;
+  method?: string;
+  url: string;
+};
+
+export type ResponsesToolStrictnessInput = {
+  targetProviderConfig?: {
+    type?: string;
+  };
+  upstreamRequest: UpstreamRequest;
+};
+
+/**
+ * OpenAI Responses treats omitted `strict` as "try to make this schema strict".
+ * Claude Code tools leave `strict` unset, so optional fields get forced on.
+ * Explicit boolean strictness is left alone. Other protocols pass through.
+ */
+export function applyResponsesToolStrictness(input: ResponsesToolStrictnessInput): UpstreamRequest {
+  const upstreamRequest = input.upstreamRequest;
+  const providerType = input.targetProviderConfig?.type?.trim().toLowerCase();
+  if (providerType !== "openai_responses") {
+    return upstreamRequest;
+  }
+  const body = upstreamRequest.body;
+  if ((upstreamRequest.bodyEncoding ?? "json") !== "json" || !isRecord(body) || !Array.isArray(body.tools)) {
+    return upstreamRequest;
+  }
+
+  const tools = stampTools(body.tools);
+  if (tools === body.tools) {
+    return upstreamRequest;
+  }
+  return {
+    ...upstreamRequest,
+    body: {
+      ...body,
+      tools
+    }
+  };
+}
+
+function stampTools(tools: unknown[]): unknown[] {
+  let changed = false;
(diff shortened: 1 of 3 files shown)


musistudio:
I think this fix is scoped a bit too broadly.

`applyResponsesToolStrictness` currently runs for every `targetProviderConfig.type === "openai_responses"` request, then stamps `strict: false` onto any function tool that omitted a boolean `strict`. That fixes the Claude Code -> Responses conversion case, but it also changes native/pass-through `/v1/responses` behavior. For a client sending an OpenAI Responses request directly through CCR, omitted `strict` is part of the API semantics; forcing it to `false` silently opts that client out of strict tool schemas.

Can we narrow this to converted Claude Code/Anthropic requests only? I’d also add a regression test like this:

```js
test("native Responses requests leave omitted strict unchanged", () => {
  const tool = monitorTool();
  const body = {
    input: [],
    model: "gpt-5.1-codex",
    stream: true,
    tools: [tool]
  };

  const input = responsesInput({
    request: {
      body,
      headers: { "content-type": "application/json" }
    },
    upstreamRequest: {
      body,
      bodyEncoding: "json",
      headers: { "content-type": "application/json" },
      method: "POST",
      url: "https://provider.example/v1/responses"
    }
  });

  const result = applyResponsesToolStrictness(input);

  assert.equal(result, input.upstreamRequest);
  assert.equal(result.body.tools[0].strict, undefined);
});

gtapps:
Sure I can do that, on it.

gtapps:
@musistudio  Pushed a follow-up (`197a8a6`) that stamps only when the inbound source is Anthropic (`sourceAdapterKey === "anthropic_messages"` or `sourceProvider === "anthropic"`). Missing source is fail-closed (leave omitted `strict` omitted).

- Native Responses (your regression, plus no source fields): identity, `strict` stays omitted
- Chat Completions -> Responses conversion: identity
- Converted Anthropic / Claude Code: omitted function `strict` still becomes `false`

Used the plugin input's source fields rather than `request.body.messages`, because Chat Completions inbound bodies also have `messages`.

PR #1729, opened 2026-08-28
fix(providers): drop Anthropic metadata on Codex and Grok CLI Responses

### LINKED ISSUE #1730: Claude Code metadata is re-injected onto openai_responses providers that reject it
## Summary

Claude Code always sends Anthropic Messages `metadata` (typically `user_id`). CCR's protocol converter drops that field, then `ccr-responses-session-affinity` copies `metadata.user_id` (and `prompt_cache_key`) onto **every** outbound `openai_responses` body except Codex backends.

Any Responses proxy with a strict schema then returns 400. This is not a Grok-only bug. Codex already hit it; Grok CLI hits it now; the next `openai_responses` preset will hit it too.

## Repro (Grok CLI)

Against CCR `3.0.22` / current main, Claude Code `2.1.250`, Grok CLI `1.0.5` (`cli-chat-proxy.grok.com`):

- `/v1/messages` **without** `metadata` → 200
- `/v1/messages` **with** `metadata` (what Claude Code sends) → 400

```json
{
  "error": {
    "message": "All target providers failed.",
    "target_provider_names": ["grok-cli-api::openai_responses"],
    "attempts": [{
      "provider_name": "grok-cli-api::openai_responses",
      "stage": "upstream_response",
      "status": 400,
      "details": { "code": "400", "error": "Argument not supported: metadata" }
    }]
  }
}
```

Codex CLI `0.150.1` reported the same class of error as `Unsupported parameter: metadata` before #1715.

## Why #1715 is not a complete fix

#1691 added session-affinity injection so multi-channel `api.openai.com` Responses can pin a conversation.

#1715 correctly **skipped** that injection for `chatgpt.com/backend-api/codex`, because the Codex backend rejects unknown body fields.

Grok CLI uses `openai_responses` against `cli-chat-proxy.grok.com`, which is **not** a Codex URL, so affinity still injects `metadata`. Growing a denylist (`isCodexResponsesUpstream`, then Grok, then Kimi, …) will lag every new strict proxy.

## Expected

Do not forward or re-inject Anthropic `metadata` unless the target is known to accept OpenAI Responses `metadata` (public `api.openai.com`). Strict subscription proxies (Codex CLI, Grok CLI, and future presets) should not see that field.

`prompt_cache_key` may need the same default: Grok was not tested for it separately because `metadata` failed first.

## Environment

- CCR `3.0.22` / current `main`
- Claude Code `2.1.250`
- Codex CLI `0.150.1`
- Grok CLI `1.0.5`

## Related

- #1688 / #1691 — session-affinity injection
- #1715 — Codex-only skip (merged)
- #1729 — proposed fix; can be expanded from per-preset `bodyRemove` to an allowlist at the affinity layer

StatPan:
Fixes #1730

## Problem

Claude Code always sends Anthropic Messages `metadata`. CCR's converter drops it, then session-affinity copies `metadata.user_id` onto **every** `openai_responses` body except Codex.

That is an Anthropic field. The public OpenAI Responses API documents `metadata`. Strict proxies do not. Grok CLI 1.0.5 returns `Argument not supported: metadata`. Codex already 400'd the same way before #1715. The next CLI preset will too if we keep a per-host denylist.

Checked live on Grok:

| Request | Result |
| --- | --- |
| no `metadata` | 200 |
| session header only (`prompt_cache_key`) | 200 |
| `metadata` present | 400 |

## Fix

One rule for all `openai_responses` providers, not a Grok special case:

- Copy Anthropic `metadata` only to `api.openai.com`
- Keep `prompt_cache_key` on every non-Codex Responses upstream (Grok accepts it)
- Codex: still skip all affinity (#1715)
- Codex/Grok `bodyRemove` still drops leftover `metadata` on the Responses body

## Test plan

- [x] `api.openai.com` gets `metadata` and `prompt_cache_key`
- [x] Grok CLI and generic Responses hosts get `prompt_cache_key` without `metadata`
- [x] Codex gets neither
- [x] Live Grok: `prompt_cache_key` 200, `metadata` 400

### DIFF
--- a/packages/core/src/agents/local-providers/codex.ts
+++ b/packages/core/src/agents/local-providers/codex.ts
@@ -614,7 +614,7 @@ function codexOauthPlugin(suffix: string, providerName = providerNamePlaceholder
 
 function codexBackendRequestTransform(): Record<string, unknown> {
   return {
-    bodyRemove: ["max_output_tokens"]
+    bodyRemove: ["max_output_tokens", "metadata"]
   };
 }
 
--- a/packages/core/src/agents/local-providers/grok.ts
+++ b/packages/core/src/agents/local-providers/grok.ts
@@ -651,6 +651,7 @@ function grokOauthPlugin(suffix: string, token: string, providerName?: string): 
   return {
     ...bearerAuthPlugin(suffix, token, {}, providerName),
     request: {
+      bodyRemove: ["metadata"],
       headers: {
         "x-grok-client-identifier": "xai-grok-cli",
         "x-grok-client-version": grokClientVersion(),
--- a/packages/core/src/gateway/core-runtime/config-compiler.ts
+++ b/packages/core/src/gateway/core-runtime/config-compiler.ts
@@ -568,6 +568,7 @@ async function withGrokOauthRuntimeDefaults(providerPlugins: unknown[]): Promise
           ...currentBodyRemove
             .map((value) => stringValue(value))
             .filter((value): value is string => Boolean(value)),
+          "metadata",
           "external_web_access"
         ]),
         headers: {
@@ -813,7 +814,7 @@ function withCodexBackendRequestTransform(request: unknown): Record<string, unkn
     : [];
   return {
     ...currentRequest,
-    bodyRemove: uniqueStrings([...bodyRemove, "max_output_tokens", "stop"])
+    bodyRemove: uniqueStrings([...bodyRemove, "max_output_tokens", "stop", "metadata"])
   };
 }
 
--- a/packages/core/src/gateway/core-runtime/local-agent-auth-provider-hook.ts
+++ b/packages/core/src/gateway/core-runtime/local-agent-auth-provider-hook.ts
@@ -420,7 +420,8 @@ const grokSupportedResponsesToolTypes = new Set([
 ]);
 
 const grokUnsupportedResponsesOptions = new Set([
-  "external_web_access"
+  "external_web_access",
+  "metadata"
 ]);
 
 function uniqueStrings(values: string[]): string[] {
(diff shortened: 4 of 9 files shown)


StatPan:
Opened #1730 to track this as a general `openai_responses` problem rather than a Grok-only config bug.

There is no existing PR that fully fixes it. #1715 stopped session-affinity from injecting `metadata` on **Codex URLs only**. Grok CLI (`cli-chat-proxy.grok.com`) and any other strict Responses proxy still get Claude Code's `metadata.user_id` copied back onto the outbound body, then 400.

This PR is offered as the fix for #1730. The current diff is the same `bodyRemove` sanitizer Codex already uses, applied to both Codex and Grok presets so `ccr grok` works now.

If that is too narrow, I would rather expand **this** PR than open a second one: change session-affinity to inject `metadata` only for upstreams known to accept OpenAI Responses `metadata` (`api.openai.com`), instead of growing `isCodexResponsesUpstream` / Grok / next-CLI denylists. Please say which shape you want and I will update the branch.

heum-ilgugim:
Updated this PR to do the full fix instead of waiting on a follow-up.

Root cause is session-affinity replaying Claude Code `metadata` onto every `openai_responses` provider except Codex. This branch now:

- copies `metadata` only to `api.openai.com`
- skips affinity entirely for Codex CLI and Grok CLI
- keeps the preset `bodyRemove` sanitizers as a backstop

#1730 is the tracking issue.

StatPan:
Corrected after a live check I should have done before expanding the diff.

Grok CLI 1.0.5 against `cli-chat-proxy.grok.com`:

- `metadata` → 400 `Argument not supported: metadata`
- session header only (`prompt_cache_key` injected, no `metadata`) → 200

This PR no longer skips all Grok affinity, and it no longer allowlists `metadata` to `api.openai.com` only. Grok keeps `prompt_cache_key` and drops Anthropic `metadata`. Other `openai_responses` providers are unchanged.

StatPan:
The Grok-only skip was the messy part. This PR now applies one rule to every `openai_responses` provider:

- Anthropic `metadata` is copied only to `api.openai.com`
- `prompt_cache_key` stays on non-Codex upstreams (live Grok check: 200)
- Codex affinity skip is unchanged

Grok is just a host that is not `api.openai.com`, not a third special case.

PR #1759, opened 2026-09-05
fix(gateway): fall back on provider error inside a 200 SSE stream

nnnet:
## Summary

OpenRouter and Anthropic commit `200 text/event-stream` response headers
before they know whether the upstream will accept the request. When it
does not, the failure can only be expressed in the body, as the first SSE
frame. The gateway decided fallback purely from the status line, so it
forwarded a "successful" 200 whose body was a single error frame and the
client reported an empty or malformed response with no retry.

The upstream executor now peeks the first meaningful SSE frame before
returning an attempt, skipping comments and pings. An error frame — or a
stream that ends without producing any event — is restated as HTTP 529,
which already classifies as a server error, so `retry` and `model-chain`
react without any change to the failure classifier. With no fallback
target left, the client receives a retryable 529 instead of an
unparseable success. Healthy streams are replayed byte-for-byte from the
peeked prefix.

Detection is on by default and can be turned off per fallback config via
`Router.fallback.detectStreamErrors`, exposed as a toggle in the routing
view. Configs written before this field default to enabled, so the fix
applies without a migration.

## Reproduction

Before the fix, a request like the following could land in the client
with an empty / malformed body when the upstream model rejects mid-stream:

```bash
curl -N -H 'Accept: text/event-stream' -H 'anthropic-version: 2023-06-01' \
  -H 'x-api-key: <redacted>' -H 'Content-Type: application/json' \
  -d '{"model":"<redacted>","messages":[{"role":"user","content":"hi"}],"stream":true,"max_tokens":32}' \
  http://127.0.0.1:3458/v1/messages
```

With the fix, the same request returns `529` and the configured
`retry` / `model-chain` fallback kicks in.

## Changes

- `packages/core/src/gateway/upstream/executor.ts` — peek the first
  meaningful SSE frame, restate provider-in-body errors as 529, replay
  healthy streams byte-for-byte.
- `packages/core/src/contracts/app.ts` — add
  `Router.fallback.detectStreamErrors` toggle (default `true`).
- `packages/core/src/config/default-config.ts` — set the new default.
- `packages/ui/src/pages/home/components/routing.tsx`,
  `packages/ui/src/pages/home/shared/routing.ts`,
  `packages/ui/src/pages/home/shared/i18n.tsx` — surface the toggle in
  the routing view.
- `packages/core/test/unit/gateway/upstream-executor.test.mjs` — cover
  error frames, empty streams, healthy stream passthrough, and the
  toggle off.
- `docker-compose.yml` — pin `container_name: claude-code-router` so the
  container can be addressed by a stable name.

## Test plan

- [x] Unit tests added in
  `packages/core/test/unit/gateway/upstream-executor.test.mjs`
  covering: error frame → 529, empty stream → 529, healthy stream
  byte-for-byte replay, comment/ping skip, toggle off bypasses peek.
- [x] `pnpm -F @musistudio/claude-code-router-core test` — green.
- [ ] Manual: replicate the `curl` above and confirm the response is
  `529` with the model-chain fallback.
- [ ] Manual: routing view shows the new toggle, default enabled, and
  disabling it preserves the old behavior.

## Risk

Low. The change is gated by a per-config toggle (default on) and the
healthy-path replay is byte-for-byte. Existing `retry` / `model-chain`
behavior is preserved — they now react to a new upstream signal, but
their decision logic is unchanged.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
       context: .
       dockerfile: Dockerfile
     image: claude-code-router:local
+    container_name: claude-code-router
     ports:
       # Publish only Nginx. Internal web/gateway listeners stay inside the container.
       - "3458:8080"
--- a/packages/core/src/gateway/upstream/executor.ts
+++ b/packages/core/src/gateway/upstream/executor.ts
@@ -22,6 +22,7 @@ import { retryDelayAfterNetworkError, retryDelayAfterStatus, shouldFallbackAfter
 import { claudeCodeOauthBetaHeader, claudeCodeOauthRequiredBeta, UpstreamRequestError } from "@ccr/core/gateway/internal/shared";
 import type { ApiKeyLimitUsage, ProviderCredentialRoutingTarget, UpstreamAttempt, UpstreamFailedAttempt, UpstreamFetchResult } from "@ccr/core/gateway/internal/shared";
 import type { RouteTraceObserver } from "@ccr/core/observability/route-trace";
+import { detectSseError } from "@ccr/core/observability/request-log-store";
 
 const providerCredentialSpilloverThreshold = 0.8;
 const openRouterDiscountModelHeader = "x-ccr-openrouter-discount-model";
@@ -467,12 +468,18 @@ export async function fetchUpstreamWithFallback(input: {
     releaseJsonObject(input.body);
 
     try {
-      const response = await fetchWithSystemProxy(attemptUrl, {
+      const fetched = await fetchWithSystemProxy(attemptUrl, {
         body: shouldSendBody(input.method) ? attempt.body?.toString("utf8") : undefined,
         headers: upstreamHeaders,
         method: input.method,
         signal: input.signal
       });
+
+      // A provider that already committed `200 text/event-stream` headers can
+      // only report a late failure inside the body. Judge the attempt on the
+      // first frame that actually arrived rather than on the status line alone;
+      // the peek restates such a response as 529 so the machinery below reacts.
+      const { response, streamError } = await peekUpstreamStreamFailure(fetched, input.fallback, input.signal);
 
       if (hasNextAttempt && shouldFallbackAfterStatus(response.status, fallbackMode)) {
         const delayMs = retryDelayAfterStatus(response.headers, failedAttempts.length);
@@ -482,7 +489,7 @@ export async function fetchUpstreamWithFallback(input: {
           kind: "outcome",
           name: "upstream.attempt.outcome",
           outcome: {
-            fallbackReason: `http:${response.status}`,
+            fallbackReason: streamError === undefined ? `http:${response.status}` : `stream-error:${streamError}`,
             retryDelayMs: delayMs,
             statusCode: response.status
           },
@@ -498,6 +505,7 @@ export async function fetchUpstreamWithFallback(input: {
           credentialChain: attempt.credentialChain,
           credentialIds: attempt.credentialIds,
           delayMs,
+          ...(streamError === undefined ? {} : { error: streamError }),
           model: attempt.model,
           statusCode: response.status
         });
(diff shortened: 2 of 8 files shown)


packages/core/src/gateway/upstream/executor.ts
```diff
@@ -1164,6 +1172,207 @@ function isRouteTraceChange(value: RequestRouteTraceChange | undefined): value i
...
+  const decoder = new TextDecoder("utf8");
+  const buffered: Uint8Array[] = [];
+  let bufferedBytes = 0;
+  let prefix = "";
+  let streamError: string | undefined;
+  let settled = false;
+
+  try {
+    while (!settled) {
+      const { done, value } = await reader.read();
```
musistudio, line 1233: `reader.read()` 这里没有独立的 peek deadline。`STREAM_PEEK_MAX_BYTES` 只在已经收到字节后才生效；如果 upstream 已经返回 200 headers，但正文迟迟没有第一个 meaningful SSE event，CCR 会卡在 attempt 判定阶段，既不把 headers/body 转发给客户端，也不进入 fallback。这个默认开启后会影响正常慢首帧流，建议加一个很短的 peek timeout，超时后按旧行为转发未判定流，或者明确把 peek timeout 当成失败策略处理。

packages/core/src/gateway/upstream/executor.ts
```diff
@@ -1164,6 +1172,207 @@ function isRouteTraceChange(value: RequestRouteTraceChange | undefined): value i
...
+      if (done) {
+        // The stream closed before a complete event. An empty body is the
+        // failure clients report as a malformed response; a partial-but-present
+        // frame is forwarded rather than second-guessed.
+        const trailing = prefix + decoder.decode();
```
musistudio, line 1238: 这里 EOF 时只把纯空白流视为空流失败，但如果 upstream 发了 `: keepalive\n\n` 或 `event: ping\n\n` 后关闭，`trailing.trim().length > 0`，于是会被当成 HTTP 200 成功转发，fallback 不会触发。PR 描述里说“ends without producing any event”会 restate 为 529；实现上最好记录是否见过 meaningful event，EOF 时只有 comment/ping 也按 failed attempt 处理，并补一个测试。

packages/core/src/gateway/upstream/executor.ts
```diff
@@ -22,6 +22,7 @@ import { retryDelayAfterNetworkError, retryDelayAfterStatus, shouldFallbackAfter
 import { claudeCodeOauthBetaHeader, claudeCodeOauthRequiredBeta, UpstreamRequestError } from "@ccr/core/gateway/internal/shared";
 import type { ApiKeyLimitUsage, ProviderCredentialRoutingTarget, UpstreamAttempt, UpstreamFailedAttempt, UpstreamFetchResult } from "@ccr/core/gateway/internal/shared";
 import type { RouteTraceObserver } from "@ccr/core/observability/route-trace";
+import { detectSseError } from "@ccr/core/observability/request-log-store";
```
musistudio, line 25: 为了复用 `detectSseError`，upstream executor 现在反向依赖了 `observability/request-log-store`。这个模块顶部会静态引入 sqlite/storage 相关依赖，把日志存储层拖进 gateway upstream hot path。建议把 SSE error detector 抽成纯模块，比如 `gateway/internal/sse-error.ts` 或 `observability/sse-error.ts`，让日志和 executor 都依赖这个纯函数。

docker-compose.yml
```diff
@@ -4,6 +4,7 @@ services:
       context: .
       dockerfile: Dockerfile
     image: claude-code-router:local
+    container_name: claude-code-router
```
musistudio, line 7: 这个改动和 SSE fallback 修复无关，建议拆成单独 PR。固定 `container_name` 还会破坏 compose project 隔离，同一台机器上两个 checkout / 两个 compose project 不能同时启动，也不能 scale 这个 service。

musistudio (commented):
整体方向我是认可的：上游在已经提交 `200 text/event-stream` 后，只能把 provider 拒绝/限流放进 SSE body，CCR 如果只按 status 判定 attempt，确实会错过 fallback。\n\n不过我建议合并前先处理下面几个点，尤其是 peek 没有独立 timeout 这一项。长期归属上，我更倾向把 provider SSE 首帧错误规范化放到 `@the-next-ai/ai-gateway`：它是最早接触 provider 协议的边界，能把 in-stream provider error 转成非 2xx/结构化失败，让所有调用方受益；但 CCR 这层仍然需要看到明确的失败信号，才能驱动自己的 `retry` / `model-chain`。如果 ai-gateway 暂时不能发版，CCR 这里做临时防护可以接受，但建议补 timeout、keepalive-only EOF、以及抽出纯 SSE error detector。\n\n我本地验证了定向用例：`node --test .test-dist/core/test/gateway/upstream-executor.test.js` 通过，14 passed。全量 core unit 在当前沙箱里有 unrelated 的 `listen EPERM 127.0.0.1` 环境失败，所以没有把它计入本 PR 问题。

PR #1763, opened 2026-09-06
fix: provider model list not loading when editing a Claude Code provider

diogomcd:
Summary

When adding a new local agent provider (Claude Code), the model list loads and models can be selected. When editing that same provider afterwards, the model catalog comes back empty. This PR fixes two gaps that combine to cause it.

## Root cause

Two independent defects break the edit flow:

1. The auto-probe effect in the provider form (`App.tsx`) called `probeProviderCandidates` without resolving provider plugins. Refreshing models manually (`refreshProviderModels`) already resolved them via `providerConnectivityProviderPlugins`, but the auto-probe did not. An imported local agent provider stores only the sentinel key `ccr-local-agent-login` as its API key; without the plugins, the probe sends that sentinel to `https://api.anthropic.com/v1/models` and gets a 401.
  
2. Even with the plugins attached, the probe still failed: saved Claude Code OAuth plugins carry the access token from import time, and the Claude Code CLI rotates that token every few hours. The gateway refreshes these plugins at config compile time (`withClaudeCodeOauthRuntimeDefaults`), but the probe path never did, so it authenticated with a stale token and got a 401.
  

## Changes

- `packages/ui/src/pages/home/App.tsx`: the auto-probe effect now resolves provider plugins with `providerConnectivityProviderPlugins(providerDraft, draftConfig.providerPlugins, existingProvider)`, mirroring `refreshProviderModels`. The effect dependency list was updated for the newly read values (`draftConfig.Providers`, `draftConfig.providerPlugins`, `providerDraft.name`, `providerEditIndex`).
- `packages/core/src/providers/probe.ts`: added `providerProbeLiveClaudeCodeOauth`, mirroring the existing Codex mechanism (`providerProbeLiveCodexOauth`). When a probe request uses the `ccr-local-agent-login` sentinel and a `claude-code-oauth` plugin is present, the probe injects the live access token read from the local Claude Code login (`readClaudeCodeOauth()`), removes `x-api-key`, and ensures the `anthropic-beta: oauth-2025-04-20` header via the existing `mergeAnthropicBetaValues` helper.
- `packages/core/test/unit/providers/provider-probe.test.mjs`: new test covering a saved plugin with a stale token plus a live local login; asserts the probe authenticates with the live token and discovers models.

## Testing

- `npm run typecheck` passes.
- `npm run test:core`: no new failures (the only failures are pre-existing Claude Design and RequestLogStore heap ones, identical with and without this change).
- Verified end to end against the real Anthropic API: with a rotated local token, `GET /v1/models` returns 200 and the full model catalog (11 models) instead of 401.

### DIFF
--- a/packages/core/src/providers/probe.ts
+++ b/packages/core/src/providers/probe.ts
@@ -13,9 +13,12 @@ import type {
   GatewayProviderProtocol
 } from "@ccr/core/contracts/app";
 import { codexDefaultBaseUrl, readCodexAuth } from "@ccr/core/agents/local-providers/codex";
+import { readClaudeCodeOauth } from "@ccr/core/agents/local-providers/claude-code";
 import { localAgentProviderApiKey } from "@ccr/core/agents/local-providers/shared";
+import { claudeCodeOauthBetaHeader, claudeCodeOauthRequiredBeta } from "@ccr/core/gateway/internal/shared";
 import { findProviderPresetByBaseUrl, providerApiKeySafetyIssue } from "@ccr/core/providers/presets/index";
 import { getProviderCatalogModels } from "@ccr/core/providers/model-catalog";
+import { isLocalClaudeCodeOauthProviderPlugin, mergeAnthropicBetaValues } from "@ccr/core/providers/oauth-plugin";
 import { fetchWithSystemProxy } from "@ccr/core/proxy/system-proxy-fetch";
 import {
   compactProviderUrl,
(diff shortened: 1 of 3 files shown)


musistudio (commented):
Thanks for the fix. I reviewed the UI and core probe paths locally.

The requirement looks valid: editing an imported Claude Code provider currently needs the same provider plugin resolution that manual model refresh already uses, otherwise the probe can send the `ccr-local-agent-login` sentinel as if it were a real Anthropic key. Refreshing the Claude Code OAuth token for the probe path also makes sense because the regular gateway runtime already refreshes local-agent OAuth credentials before traffic is sent.

I do not think the primary fix belongs in `@the-next-ai/ai-gateway` as-is. The broken path here is CCR management UI direct probing of upstream provider endpoints, not normal requests through the ai-gateway runtime. Unless this probe is changed to go through the gateway `/v1/models` endpoint or ai-gateway exposes a reusable provider-hook request builder, an upstream-only change would not fix the edit form model list.

Two things to address before merge:

1. This PR is now stale against current `main`. A local merge check against `origin/main` shows conflicts in `packages/core/src/providers/probe.ts`, `packages/core/test/unit/providers/provider-probe.test.mjs`, and `packages/ui/src/pages/home/App.tsx`. Please rebase and keep the newer main changes, especially the recent provider probe additions.
2. The new Claude Code live-token logic in `probe.ts` duplicates behavior that already exists in the local-agent auth provider hook. It is okay for this fix, but it would be cleaner to share a small CCR helper so the gateway runtime and probe path stay in sync if the required Anthropic OAuth headers change again.

I ran the targeted provider probe test after building core unit tests. The provider probe suite passed locally.

Issue #1783, opened 2026-09-10
responses: 400 'array too long' when history contains a reasoning item with non-empty content


Burlesque1:
## Summary

Codex API (`openai_responses`) requests 400 with `All target providers failed` when the translated request contains a `reasoning` item whose `content` array is non-empty. The upstream rejects it because the Responses API requires a `reasoning` item's `content` to be empty (max length 0).

## Upstream error

```
Invalid 'input[39].content': array too long.
Expected an array with maximum length 0, but got an array with length 1 instead.
type: invalid_request_error
param: input[39].content
code: array_above_max_length
```

## Reproduction

Request `model: "gpt-5.6-luna"` (Codex API profile) with streaming. The offending item is index 39 in the `input` array:

```json
{
  "type": "reasoning",
  "id": "rs_9a5d56e5879d4fb5855a9a7c7783f75c",
  "summary": [],
  "content": [
    { "type": "reasoning_text", "text": "We need respond, need clarify what content wrong..." }
  ],
  "encrypted_content": null
}
```

The `content` field holds a single `reasoning_text` block, which this Responses API endpoint does not allow on a `reasoning` input item (it only accepts `summary` and/or `encrypted_content`). Any history containing a translated Claude thinking block produces this and 400s deterministically.

## Root cause — lives in `@the-next-ai/ai-gateway` (NOT this repo)

The OpenAI Responses translation emits reasoning text into the `content` array of `reasoning` input items. The culprit is `buildOpenAIResponsesReasoningInputItem` in:

**`The-NeXT-AI/ai-gateway` → `src/adapters/builtins/target/openai-responses.ts` (line ~694)**

```ts
...(reasoning.text
  ? {
      content: [
        { type: 'reasoning_text', text: reasoning.text }
      ]
    }
  : {}),
```

CCR just consumes this package (`@the-next-ai/ai-gateway@^1.0.18`). A proper root fix belongs upstream there — e.g. fold `reasoning.text` into `summary` (which the endpoint accepts) rather than emitting a non-empty `content`.

[...]

PR #1784, opened 2026-09-10
fix(openai): strip non-empty content from Responses reasoning input items

### LINKED ISSUE #1686: Anthropic thinking content blocks in message history break openai_responses upstream (400) - stripping them makes the same body pass

Burlesque1:
## What

OpenAI **Responses** upstreams (e.g. Codex API) reject a `reasoning` input item whose `content` array is non-empty with HTTP 400 `array_above_max_length`. Any request whose history contains a translated Claude thinking block carries a `reasoning` item with `content: [{ type: "reasoning_text", ... }]`, so the request 400s deterministically.

This extends `stripUnsupportedOpenAiRequestParameters` to neutralize that, complementing the existing message-history `thinking`-block strip.

This supersedes #1702 (the `messages[]` thinking-block strip) — both changes live in the same function and this PR carries both, so it fixes **#1686 and #1783** together.

## Change

`packages/core/src/gateway/upstream/executor.ts`:

- `stripThinkingContentBlocks(messages)` — strips `type: "thinking"` / `type: "redacted_thinking"` blocks from `messages[].content[]` (from #1702).
- `stripResponsesReasoningContent(input)` — for the Responses `input` array:
  - drops `content` from each `type === "reasoning"` item,
  - drops empty `summary: []`,
  - drops the item entirely when nothing accepted by the endpoint remains (`summary` / `encrypted_content`).
- `stripUnsupportedOpenAiRequestParameters` runs both alongside the existing top-level `thinking`/`reasoning_split` strip.

`packages/core/test/unit/gateway/upstream/executor.test.mjs`: covers thinking/redacted_thinking stripping, reasoning `content` drop, empty-reasoning-item removal, items retaining `summary`/`encrypted_content`, and no-op passthrough.

## Notes

- Defensive belt-and-suspenders: the root fix belongs upstream in `The-NeXT-AI/ai-gateway` (`buildOpenAIResponsesReasoningInputItem`); this repo uses `@the-next-ai/ai-gateway@^1.0.21` and cannot control its release timing.
- Related: fixes #1686 and #1783. Root-cause upstream PR: The-NeXT-AI/ai-gateway#29.


musistudio:
I think this needs one small fix before merge.

OpenAI Responses reasoning input items still require `summary`, even when it is an empty array. For the reported bad shape (`summary: [], content: [...], encrypted_content: null`), the new sanitizer removes `summary` and then keeps the item because the `encrypted_content` property is present, leaving something like:

```json
{ "type": "reasoning", "id": "...", "encrypted_content": null }
```

That can still fail validation, just with a missing required `summary` instead of `content` being too long. I think the safer behavior is to preserve `summary: []` while stripping `content`, or only keep a summary-less item when `encrypted_content` is a valid string; otherwise drop the reasoning item.

Burlesque1:
Addressed the review feedback in 13dddf6b:

- `stripResponsesReasoningContent` now **preserves `summary`** (including an empty array) instead of deleting it when stripping `content`. An empty `summary: []` is a valid field, so the reasoning item stays valid.
- It only keeps a **summary-less** item when `encrypted_content` is a valid non-empty string; otherwise the item is dropped.

For the reported bad shape (`summary: [], content: [...], encrypted_content: null`), the result is now `{ type: "reasoning", ..., summary: [] }` — content removed, summary preserved, no missing-required-field failure.

Updated the test assertion accordingly (empty summary now expected to be preserved) and verified 5 scenarios: the maintainer-reported shape, drop-when-only-rejected-content-remains, valid-encrypted keep, non-empty summary keep, and non-reasoning passthrough.

musistudio:
Is it more reasonable to use this PR directly?
https://github.com/The-NeXT-AI/ai-gateway/pull/29

PR #1788, opened 2026-09-12
fix(claude-code): scope subagent routing elements to real subagent launches; record uncaptured upstream status

Aser-Osama:
## What this fixes

Three independent defects found while instrumenting subagent routing. All are in the
gateway/observability path; none touches the OpenCode Go feature in #1786.

### 1. A quoted routing element could reroute the request that was reviewing it

`CCR-SUBAGENT-MODEL` in a request's first two user messages was honored as a routing
directive regardless of what kind of request carried it. The security-classifier request
embeds the transcript it is judging in its first user message, so a transcript that merely
*mentioned* an element silently rerouted the classifier to whichever model the reviewed
text named — and stripped the element, denying the classifier the evidence. A parent
conversation quoting an element was exposed the same way.

**Reproduced**, two ways. In a local database, classifier requests tracked whichever model
a concurrently-running child had been tagged with, each window sitting strictly inside that
child's window and carrying no agent id:

| tagged child model | child window | classifier rerouted to same model |
|---|---|---|
| model A | 22:58:05 → 23:10:13 | 22:58:10 → 23:10:10 |
| model B | 23:28:19 → 23:33:39 | 23:28:29 → 23:33:36 |
| model C | 23:52:54 → 00:08:47 | 23:53:02 → 00:08:34 |

And directly on the wire: of the captured requests carrying an element in their first two
user messages, 16 were genuine subagent launches (marker true) while 43 were classifier or
parent requests (marker false) that had no business being rerouted.

**Impact:** a request's provider/model can be chosen by text it is only quoting. For the
classifier that means its prompt — which contains a session transcript — can be sent to a
provider the operator did not choose for it.

**Fix:** the message slot is trusted only when the request declares itself a subagent launch
via the `cc_is_subagent` billing marker. The element is left in place for every other
request so a classifier still sees it.

`system` stays unconditional **on purpose**: it is author-controlled configuration (the
caller's own prompt, or a tag injected from the profile / user config), and honoring an
element there is the documented way to force a model. Gating it would break that documented
behavior and four existing tests. The residual risk — reviewed content arriving in `system`
rather than in a message — is recorded in a code comment; no observed client does that.
Happy to gate it too if you would rather trade the documented mechanism for the stricter
boundary.

**Validated after the fix, on live traffic from a patched build.** The requests below all
carried an element in the gated slot — the first two user messages — with `system`
uninvolved, so the marker gate is what separates them:

| declares `cc_is_subagent` | requests | outcome |
|---|---|---|
| yes | 56 | routed as a subagent |
| no | 15 | **not rerouted** — stayed on the parent model, `route_reason = default` |

The 15 are parent-conversation requests whose first user message quotes an element (a
context-compaction summary of the work that produced these elements). That is exactly the
quoting scenario above, and before this change all 71 would have been rerouted by text they
were only carrying.

The marker is read by a new non-mutating helper that accepts any `system` shape. Deriving
it from `removeClaudeCodeBillingSystemHeader` alone was too narrow — that function only
reports the marker for a strippable block at `system[0]`, so a child whose system is a
plain string, or whose billing block was reordered by a proxy, would have silently lost
routing and forwarded the element upstream. Both shapes are already pinned as real inputs
by existing tests.

### 2. An uncaptured upstream status was indistinguishable from a failure

The raw-trace producer omits `statusCode` on several provider paths, serializing its
response metadata to `{}`. That became `status_code = 0` with `ok = 0` — the same shape as
a real failure. In a local database, 534 of 757 rows were affected, including requests that
demonstrably succeeded and whose responses were used.

**Fix:** a private `upstream_outcome` column records how the attempt actually ended:
`http_status` | `unknown` | `stream_failure` | `transport_failure` | `cancelled`. A real
HTTP status is always preserved and **a 200 is never synthesized**.

**Measured on a live database after the fix** (3962 rows):

| | rows |
|---|---|
| counted as errors before | 2637 (67%) |
| counted as errors after | 396 |
| reclassified as undecided | 2241 — every one `status_code = 0` with outcome `unknown` |
| genuine failures retained | 295 × `429`, 101 × `401`, all `http_status` |

No real failure was reclassified: every retained row kept a true HTTP status, and every
reclassified row is one where the producer never reported a status at all.

Every consumer that reported failures now agrees:

- the request-log error filter excludes a row that is `unknown` with no error;
- the agent error list uses the same predicate;
- an agent trace run reports `partial` for an undecided row rather than `error`, reusing the
  status value the contract already had, so no consumer had to change. One cosmetic
  consequence, flagged rather than hidden: the dashboard already renders that state as
  "Partial failure" (`dashboard.tsx:5992`), which reads oddly for a request that is merely
  undecided. Renaming it would also relabel genuine partial-failure runs, so it is left
  alone here; a follow-up could split the label without touching the union;
- agent totals count only decided requests, so an undecided row no longer inflates
  `errorCount` or depresses `successRate`;
- usage events carry the outcome, so Overview totals stop counting an undecided row as an
  error. This needed a second fix: the column was written only by the request-log backfill,
  which produces **no rows at all** on a single-gateway install, so every usage row came from
  `UsageStore.record`, which never wrote it. On the live database that left 2241 of 3952 rows
  at `status_code = 0` with an empty outcome — still counted as errors, i.e. 57% of Overview's
  error count was phantom. `record()` now writes the outcome it can establish from the status,
  and rows left empty by an earlier build are migrated on open;
- the logs UI paints those rows neutrally instead of red.

Subagent attribution is now anchored to the start of the text. The router deliberately
ignores an element that is merely quoted, so scanning anywhere in a payload invented a child
call named after the quoted text and showed a phantom subagent row. A child prompt puts the
element first, which is also what the gateway requires before it will route.

### 3. A reroute left no record of what the client asked for

`model` and `resolved_model` both hold the target that was *used*, and the legacy
`requested_model` is populated after routing, so it holds the used target too. The only
place the original ask survived was the raw-trace bundle's request headers — and those are
deleted seconds later when the spool is cleaned up. After that, nothing distinguished a
request that was rerouted from one that asked for the model it got.

**Fix:** `client_model`, `route_reason` and `route_source` are persisted on the row. Only the
model string is read from the client body; no prompt or credential is copied, and a test
asserts both.

The logs list now reads them, which is the surface the columns exist for: its "asked -> used"
pair came from `requestedModel`, so on exactly the requests where the comparison matters it
rendered the same model on both sides. The dashboard's route-reason column also stops
depending on headers that no longer exist once the bundle is gone.

This **is** verified end to end on live traffic. Getting there took one more
fix, and the diagnosis in commit `1e8b728`'s own message was wrong — `af3190b` corrects it.

The columns were empty on every row of a live database not because the raw-trace update
fails to land, but because on a single-gateway runtime that update is the *only* producer.
When no row exists yet for a request, the runtime converts the bundle straight into a record
input via `standaloneRecordInputFromRawTrace` instead of updating a gateway row, and that
conversion is a hand-written field-by-field map which never copied `clientModel`,
`routeReason` or `routeSource`. The reader extracted them correctly, the update path assigned
them correctly, the INSERT bound them correctly — and they were dropped in between.

The same map is why `upstream_outcome` worked on this very path while these three columns did
not: it already copied `upstreamResponseReceived` and `upstreamCancelled`. One mapped field
sitting beside three unmapped ones is what made the earlier diagnosis look plausible.

Corroborating evidence from the live database: every row carries
`event_id = 'raw-trace:<id>'`, and `request_route_traces` holds **0** rows — so `record()`,
which inserts a route trace whenever the gateway supplies one, had never run there at all.

Verified on live traffic after the fix:

| row | `client_model` (asked) | `model` (used) | `route_reason` | `route_source` |
|---|---|---|---|---|
| parent | `…/claude-opus-5` | `claude-opus-5` | `default` | `default` |
| child | `…/claude-opus-5` | a routed worker model | `builtin:claude-code-subagent` | `subagent` |

The child row is the case that was previously unreadable: `requested_model` and
`resolved_model` both hold the post-routing target, so without `client_model` there is no way
to see the request was rerouted at all.

The `1e8b728` change is kept — it is what populates these columns for a request recorded by
the gateway path rather than from a bundle.

One adjacent gap this exposed is **left alone deliberately**: on a standalone insert
`writeBatch` returns before the follow-up raw-trace UPDATE, so the `gateway_*` columns and
`request_route_traces` stay empty in this topology. That predates this PR and is not routing
evidence.

## Compatibility

- `status_code`, `ok`, `error` and the legacy `requested_model` keep their exact current
  meaning. `requested_model` remains post-routing; `client_model` is the new field that
  exposes the original ask.
- `upstream_outcome` is additive, defaulting to `unknown`. `partial` already existed in
  `AgentAnalysisTraceRunStatus`, so no public union was widened.
- Outcome evidence merges monotonically within one attempt: a stored `cancelled` is
  preserved, and a stored `stream_failure` is preserved unless incoming evidence is
  `cancelled`. Without that, a second bundle for the same request and the same attempt —
  which context-archive continuation can produce — could downgrade a recorded stream failure
  to `http_status` and flip `ok` back to true while leaving stale error text. Incoming
  cancellation is computed from incoming evidence only, so a status-only update cannot
  masquerade as a cancellation.
- A bare upstream 499 stays `http_status`; only the gateway's own client-disconnect marker
  means `cancelled`.

## Schema and backfill

`upstream_outcome`, `client_model`, `route_reason`, `route_source` are added to fresh
schemas and to existing `request_logs` databases by idempotent `ALTER` statements; the
outcome column's ALTER plus backfill runs in one transaction, so an interrupted upgrade is
retried on next open. `usage_events` gains `upstream_outcome` through the same guarded
pattern already used there, plus an idempotent migration that fills rows left empty by a
build which added the column without populating it (positive status → `http_status`, zero
status → `unknown`). Backfill marks rows with a positive status `http_status` and
rows carrying the disconnect marker `cancelled`, leaving legacy zero-status rows `unknown`.

That last rule was challenged in review, on the theory that a zero-status row with an error
string is really a `transport_failure`. Checked against the live database instead of argued:
of its 2241 zero-status rows, **0 carry any error string and all 2241 have a stored response
body** — every one demonstrably received an upstream response and simply had no status
reported. `unknown` is the accurate label for them, not merely the cautious one, and the
proposed heuristic would have relabelled nothing while risking invented errors.

The `usage_events` backfill is **paged** (500 rows per batch) rather than one unbounded
`UPDATE`. It runs while the usage database is opening, and on a large history a single
statement would hold the write lock long enough to push concurrent writers past
`busy_timeout` — and the usage capture path swallows those failures, so requests would have
vanished from Overview rather than erroring visibly. No progress ledger is needed: the
predicate is its own progress marker, since every row the update touches stops matching it,
so the loop is idempotent, resumable after a crash mid-migration, and self-terminating.

Both new `ALTER TABLE ADD COLUMN` statements read `PRAGMA table_info` and then add the
column, so a second opener (the request-log worker thread, another app instance, a CLI
invocation) can win that race in between and the loser would abort the database open with
"duplicate column name" on the first launch after upgrading. Both now treat that specific
error as the state they wanted and continue to the idempotent backfill.

The `request_logs` column add and its backfill share one `BEGIN IMMEDIATE`. SQLite applies
DDL transactionally, so a crash between them rolls the column back and the next open redoes
both — the column-presence guard cannot be true unless the backfill also committed. Checked
rather than assumed, by killing a process mid-transaction after the `ALTER` and confirming
the column was absent on reopen.

Retention is unaffected: it deletes whole rows, so the new columns need no cleanup of their
own, and a test pins that.

No migration-ledger row was added: an existing integration test pins that ledger with a deep
equality assertion, and the transactional column presence is itself an idempotent completion
marker.

## Tests

`node build/run-tests.mjs core` on this branch: **995 tests, 989 pass, 1 fail, 5 skipped**.
UI 209/209, electron 44/44, architecture 5/5, CLI 5/5, `npx tsc --noEmit` clean — every
suite in the repo, 1258 tests total.

All figures are measured in a clean worktree checked out at this branch, not in a development
tree carrying other work. That distinction matters: a tree that also held an unrelated feature
branch reported 1011 core tests and 209 -> 211 UI tests, because it contained test files this
branch does not.

**The single failure is pre-existing on `main` and unrelated to this PR**: `Claude Code local
provider prefers macOS Keychain credentials over stale file credentials`. That file is
identical on `main`, is not modified here, and is not in this diff — it appears
environment-dependent. I am explicitly **not** claiming a fully green suite.

New coverage: the observed classifier request shape, a quoted element in a parent
conversation, a native-workflow child, both widened marker shapes, an omitted status with a
tool continuation, a preserved 429, stream failure and cancellation after headers, transport
failure before any response, the out-of-order downgrade, bare 499 with and without the
disconnect marker, reopen idempotency, undecided rows in agent analysis and in usage totals,
a quoted element earning no subagent attribution, and route evidence upgrading an existing
database, surviving a reopen, and being pruned wholesale by retention, and the standalone
raw-trace branch carrying routing evidence through to the stored columns. The boundary tests and
the marker-shape tests were each mutation-checked — they fail with the fix reverted.

## Evidence provenance

- The routing defect was reproduced against an installed 3.1.0 build that **predates** this
  patch. That evidence demonstrates the *defect*, not the fix.
- The fix is verified by the local suite above **and** by the post-fix live table in section
  1, observed on a build carrying this patch: element-bearing requests that do not declare a
  subagent are no longer rerouted, while genuine launches still are.
- The route-evidence columns are verified on live traffic after the final fix; the
  `upstream_outcome` column likewise, including a real upstream 401 correctly retained as
  `http_status` rather than masked as undecided.
- The native-workflow child request shape used in one test was observed from real
  multi-agent runs on that same pre-patch build.
- Reproductions here are sanitized. No transcripts, credentials, or provider identifiers
  from the originating environment are included; test fixtures use neutral names.

## Not addressed

This does not touch upstream rate limiting. A separate `429` observed during this work came
from the ultimate upstream with its own request ids, with zero gateway retries; no change
here affects a real provider quota.


packages/core/src/observability/request-log-store.ts
```diff
@@ -4466,6 +4677,13 @@ function ensureRequestLogSchema(database: SqlDatabase): void {
   addColumn("response_body_truncated", "INTEGER NOT NULL DEFAULT 0");
   addColumn("response_body_ref", "TEXT NOT NULL DEFAULT ''");
   addColumn("error", "TEXT NOT NULL DEFAULT ''");
+  // Routing evidence. `model`/`resolved_model` hold the target CCR actually used,
+  // so these record what the client asked for and why it moved. Legacy
+  // `requested_model` keeps its existing post-routing meaning for current callers.
+  addColumn("client_model", "TEXT NOT NULL DEFAULT ''");
```
musistudio, line 4698: These new route-evidence columns still go through the generic `addColumn()` helper, which reads `PRAGMA table_info` once and then runs a bare `ALTER TABLE`. That leaves the same first-upgrade race this PR fixed for `upstream_outcome`: another opener can add one of these columns after the PRAGMA read but before this `ALTER`, causing this process to fail opening the database with `duplicate column name`. Could we make these column adds duplicate-tolerant too, or make `addColumn()` itself tolerate that specific error and refresh/update the `columns` set?

packages/core/src/observability/raw-trace-sync.ts
```diff
@@ -1613,20 +1613,28 @@ export async function readRawTraceRequestLogBundle(
...
     clientRequestMetadata,
+    clientRequestBody,
     upstreamRequestMetadata,
     upstreamResponseMetadata,
     upstreamRequestBody,
     upstreamResponseStream,
     fallbackResponseBody
   ] = await Promise.all([
     readRawTraceJsonPart(parts, "client_request_metadata", spoolDirectory),
+    readRawTraceJsonPart(parts, "client_request", spoolDirectory),
```
musistudio: This reads and `JSON.parse`s the entire `client_request` part just to extract the `model`. Raw trace parts are intentionally allowed to be much larger than the normal request-log body limit, and the store usually moves those bodies through sidecar/limited paths instead of loading them wholesale. On large prompts or multimodal requests, replay/sync can now pull a very large client body into memory even when body capture would otherwise be bounded. Could the client model be emitted in metadata, or extracted through a bounded/streaming path instead?

musistudio (commented):
Thanks for the detailed write-up. The overall direction makes sense to me: gating message-carried subagent routing on the real Claude Code subagent marker is a good boundary, and separating an uncaptured upstream status from an actual failure is the right shape. I found two implementation issues worth addressing before merge.

packages/core/src/observability/request-log-store.ts
Aser-Osama: Done in `bec78af` — made `addColumn()` itself tolerant instead of special-casing the three columns.

`addColumnDuplicateTolerant` now backs every `request_logs` column add (all of them, not just the route-evidence three), plus the two other PRAGMA-once adds on the same open path: `request_route_traces.trace_json` and `request_log_pending_updates.update_bytes`. It adopts `duplicate column name` as the state it wanted and updates the `columns` set; every other error still throws, so a genuinely broken schema is not masked. `ensureUpstreamOutcomeColumn` keeps its own `BEGIN IMMEDIATE` + backfill handling unchanged.

Test: `request-log-upstream-outcome.test.mjs` drives the race state directly against real SQLite — a stale `columns` set with the column already present — and asserts no throw and the set updated, that a genuinely missing column is still added, and that `no such table` still propagates. Core suite: 1030 tests, 0 failures.

packages/core/src/observability/raw-trace-sync.ts
Aser-Osama: Done in `bec78af` — the `client_request` part is no longer read at all.

The gateway stamps the pre-routing ask on a trusted `x-ccr-client-model` route header, from both producers: the core-routing `beforeRouting` transform and the public pipeline at ingress (including non-routing paths, so the raw-trace path does not lose the ask where the body read used to supply it). The header is in the `ccrRouteHeaderNames` strip list, so a client-forged value is discarded before the trusted one is written. The reader decodes it from `client_request_metadata.headers` next to `x-ccr-route-reason`/`x-ccr-route-source`, so no part larger than a metadata header is loaded.

Two details worth flagging:

- The value is base64url-encoded, not passed through `sanitizeHeaderValue`. Model selectors can carry non-ASCII provider names, and an ASCII-coerced header would both lose the ask and let the raw-trace update overwrite the correct value `record()` wrote. A non-ASCII selector now round-trips verbatim.
- A bundle without the header leaves an existing `client_model` untouched (`pushValue` skips undefined), so old spool bundles cannot clear a correct value.

Tests: a decoy body model that must not be read; a non-ASCII round-trip; a header-less bundle preserving the recorded value; and both producers stamping it (routed and non-routed) with a forged inbound value replaced, asserted through the real pipeline executor. Core suite: 1030 tests, 0 failures.

Aser-Osama:
Two open issues this should fix:

- **#1801** — the `status_code = 0` rows (HTTP `-`, red, but with real token counts) are the uncaptured-status case: no longer counted or painted as failures; only real statuses stay red.
- **#1540** — `client_model`/`route_reason`/`route_source` are persisted and the log list shows asked → used, so reroutes are visible.

packages/core/src/usage/store.ts
```diff
@@ -450,7 +461,8 @@ export class UsageStore {
             logs.cache_write_tokens,
             logs.total_tokens,
             logs.cost_usd,
-            'request_log'
+            'request_log',
+            logs.upstream_outcome
```
musistudio: This backfill now assumes every attached request-log database already has the new `upstream_outcome` column. On upgrade that is not guaranteed: the Usage/Overview path can open before the request-log worker has migrated the old `request_logs` schema, so this SELECT fails with `no such column: logs.upstream_outcome` and the whole request-log backfill is skipped. Please check the attached table schema first, or fall back to `CASE WHEN logs.status_code > 0 THEN "http_status" ELSE "unknown" END` when the column is absent.

packages/core/src/usage/store.ts
```diff
@@ -728,6 +792,7 @@ const usageTotalsSelect = `
             COALESCE(SUM(COALESCE(cost_usd, 0)), 0) AS cost_usd,
             COALESCE(SUM(duration_ms), 0) AS duration_ms,
             COALESCE(SUM(CASE WHEN status_code >= 200 AND status_code < 400 THEN 1 ELSE 0 END), 0) AS success_count,
+            COALESCE(SUM(CASE WHEN status_code = 0 AND upstream_outcome = 'unknown' THEN 1 ELSE 0 END), 0) AS undecided_count,
```
musistudio, line 841: This fixes the aggregate SQL paths, but `recentRequests` still goes through the in-memory `buildTotals([event])` path below. That path does not read `upstream_outcome` into `StoredUsageEvent` and still derives errors as `requestCount - successfulRequests`, so a `status_code = 0` / unknown-outcome row still appears as a failed recent request even though these aggregate totals treat it as undecided.

packages/core/src/usage/store.ts
```diff
@@ -484,6 +496,44 @@ export function onUsageRecorded(listener: () => void): () => void {
...
+ * duration -- long enough on a large history to push concurrent writers past
+ * `busy_timeout`, and those failures are swallowed by the usage capture path, so
+ * requests would vanish from Overview instead of erroring visibly.
+ *
+ * No progress ledger is needed: the predicate is its own progress marker, since
+ * every row this touches stops matching it. So the loop is idempotent, resumable
+ * after a crash mid-migration, and self-terminating.
+ */
+function backfillUsageUpstreamOutcome(database: SqlDatabase): void {
+  const selectBatch = database.prepare(`
```
musistudio, line 541: This migration is functionally idempotent, but each batch starts the same `WHERE upstream_outcome = "" ORDER BY id ASC LIMIT 500` scan again. On a large usage history, every later batch has to scan past the already-migrated prefix, so the total read work can grow much faster than the number of rows being updated. Keeping a local `lastId` and selecting `id > lastId` for the next page would preserve the short write batches without repeatedly rescanning old rows.

packages/core/src/usage/store.ts
Aser-Osama: Done in `af8fba4`. `backfillFromAttachedRequestLog` now reads `PRAGMA request_log_source.table_info(request_logs)` before building the insert. If `upstream_outcome` is there it uses `logs.upstream_outcome`. If not, it falls back to `CASE WHEN logs.status_code > 0 THEN 'http_status' ELSE 'unknown' END`, which is the same rule the request-log migration uses to seed the column. So an attached database that has not been migrated yet still backfills, and gets the same outcomes it would have after migrating.

Test: `UsageStore backfills from a request-log database that predates upstream_outcome` builds a `request_logs` table with the old schema (no `upstream_outcome`) holding a 200 row and a status-0 row. It checks that both rows are backfilled (`requestCount` 2, `errorCount` 0) and stored as `http_status` / `unknown`. Without the fix it fails, because the backfill gets skipped.

packages/core/src/usage/store.ts
Aser-Osama: Done in `af8fba4`. `readRecentRequestRows` now selects `upstream_outcome`, and `StoredUsageEvent` carries it. `buildTotals` uses the same rule as `usageTotalsFromRow`: a row with `status_code = 0` and `upstream_outcome = 'unknown'` is undecided, so it does not count as an error and is left out of the success-rate denominator (`errorCount = max(0, decided - successful)`, `successRate = decided > 0 ? successful / decided : 0`). The in-memory and SQL paths now agree.

Test: the existing `does not count an uncaptured upstream status as an error` test now checks `recentRequests` as well. The 500 row has `errorCount` 1, the 200 row has 0, and the status-0 row has 0. The last check failed before the fix.

packages/core/src/usage/store.ts
Aser-Osama: Done in `af8fba4`. The migration now keeps a local `lastId`. The select is `WHERE id > ? AND upstream_outcome = '' ORDER BY id LIMIT 500`, and the update is limited to `id > lastId AND id <= batchLastId`. Each page seeks past the rows already migrated instead of scanning them again, and the writes stay in the same short batches. It is still idempotent and resumable: a restart begins at `lastId = 0`, and rows that were already migrated no longer match `upstream_outcome = ''`.

The existing 1,250-row test (more than two pages, with mixed statuses across page boundaries) still passes and confirms that no row is skipped. Core suite: 1031 tests, 0 failures. Typecheck is clean.

musistudio (commented):
Thanks for the thorough write-up and the follow-up fixes. The subagent-routing boundary looks sound for the observed request shapes, but I found two remaining correctness issues in the Usage/Overview path.

1. **A `2xx` stream failure is still counted as a success.**

   `usageTotalsSelect` computes `success_count` from `status_code` alone (`packages/core/src/usage/store.ts:840`), and the in-memory `buildTotals` path does the same (`packages/core/src/usage/store.ts:1140`). Consequently, a row such as:

   ```text
   status_code = 200
   upstream_outcome = stream_failure
   ```

   produces `errorCount = 0` and `successRate = 1`, even though the request-log side correctly records `ok = 0` and `upstream_outcome = stream_failure`. I reproduced this with a focused regression test that inserted that exact combination: the expected `errorCount === 1` assertion failed with `0 !== 1`. A `2xx` cancellation is likewise treated as a successful request rather than a cancelled turn.

2. **The raw-trace path discards the outcome before writing the usage row.**

   `readRawTraceRequestLogBundle` already establishes whether an upstream response existed, and the request-log path can derive `transport_failure`, `stream_failure`, `cancelled`, etc. However, `recordUsageCaptureFromRawTrace` passes only `statusCode` into `recordGatewayUsageCaptureIfMissing` (`packages/core/src/observability/raw-trace-sync.ts:1523-1535`), while `UsageCaptureInput` has no outcome field. `UsageStore.record()` then maps every zero status to `unknown` and every positive status to `http_status` (`packages/core/src/usage/store.ts:223-229`).

   This means a no-response transport failure is stored as undecided, and an SSE failure after HTTP 200 is stored as an ordinary HTTP result. Because raw-trace usage capture runs before the request-log enqueue (`packages/core/src/observability/raw-trace-sync.ts:394-409`) and later usage backfill/delivery deduplicates by `requestId`, the correctly classified request-log row cannot repair the already-written usage row.

I think `UsageCaptureInput` needs to carry the resolved upstream outcome (or equivalent failure evidence), and both the SQL and in-memory aggregate predicates should define success/error/undecided from that outcome as well as the HTTP status. Regression coverage should include at least `200 + stream_failure` and `0 + transport_failure` through the public/raw-trace writer paths.

Issue #1791, opened 2026-09-12
Investigate reported zero cache usage with Claude Code via OpenCode Go


Yongzhaooo:
## Update: attribution root-cause hypothesis withdrawn

Further verification does not support the proposed attribution-header fix for the current-version custom-base-URL case. PR #1792 is being withdrawn. The originally reported OpenCode Go cache/usage symptom remains unresolved; this update does not establish a CCR defect or claim that the upstream symptom has been fixed.

The [official Claude Code gateway protocol documentation](https://code.claude.com/docs/en/llm-gateway-protocol#system-prompt-attribution-block) states:

> From Claude Code v2.1.181, the block is stable for the lifetime of a conversation when requests route through a custom base URL.

Before v2.1.181 the block contained a per-request token. The [environment-variable documentation](https://code.claude.com/docs/en/env-vars) scopes that workaround to those older versions. This does not justify defaulting `CLAUDE_CODE_ATTRIBUTION_HEADER=0` for all CCR-managed Claude Code profiles based on the original report.

## Local reproduction

On Windows, the installed Claude Code 2.1.220 CLI sent actual requests to an isolated loopback receiver returning synthetic Anthropic responses. Each of three sessions completed three successive turns using model name `deepseek-v4.1-flash`, with the attribution variable unset, set to `0`, and set to `1` respectively.

| Setting | Attribution text in JSON `system` | System content across three turns |
| --- | --- | --- |
| Unset | Present, unchanged | Identical |
| `0` | Absent | Identical |
| `1` | Present, unchanged | Identical |

The captured line was:

```text
x-anthropic-billing-header: cc_version=2.1.220.35b; cc_entrypoint=sdk-cli;
```

There was no per-turn `cch` variation. This line was prompt text inside the request body, not an HTTP header. All nine requests included `cache_control` markers despite the non-Claude model name.

These checks used print/stream-json mode with tools, hooks, and MCP disabled, isolated settings, and no persisted sessions. They did not use `--bare`, which itself skips attribution. No request was sent to OCG, and no real cache-hit or billing improvement was measured. Other client builds, modes, and production routing paths were not tested.

## Public OpenCode code check

[...]

PR #1792, opened 2026-09-12
fix(claude-code): disable attribution header by default in managed profiles

Yongzhaooo:
## Summary

Default `CLAUDE_CODE_ATTRIBUTION_HEADER` to `0` in CCR-managed Claude Code profiles, so supported Claude Code clients omit the attribution line from their system prompt. This removes one possible source of prefix variation; it does not guarantee upstream cache hits.

Refs #1791. The broader cache-miss and usage-reporting investigation remains open.

## Behavior and scope

- Apply the default to new profiles and existing profiles when normalized if the variable is absent. Preserve explicit profile overrides, including `1` to re-enable attribution.
- Use a profile-wide default, including profiles routed to Anthropic. Profiles may change models or route dynamically, so this change does not infer the final provider from the selected model name.
- Carry the value into generated Claude settings and launcher environment through the existing profile environment machinery. Track it as a managed Claude settings key and remove it when converting to another agent.
- Make the UI's create/edit/save and agent-switch behavior consistent with the backend.

No OpenCode Go protocol, usage mapping, session-affinity, or provider-billing behavior is changed.

## Validation

Windows, Node 24.19.0; reviewed base `afba89fa3778c772574636f1bd9aaad9138b44b7`.

- TypeScript: `node node_modules/typescript/bin/tsc --noEmit` passed.
- Test compilation: `node build/test.mjs core ui` passed.
- Entire compiled UI suite: 210 passed, 0 failed.
- Profile service integration tests plus UI profile component tests: 66 tests, 62 passed, 1 skipped, 3 failed. All three failures reproduce on the unchanged base with the same Windows runtime: stale legacy profile flags, refreshing managed global Claude gateway env, and restoring global settings with only CCR-scoped profiles. They involve user-home/global-settings handling.
- The three added tests passed. They cover loading legacy profiles without the variable, preserving an explicit opt-in, applying the value to generated settings, UI create/edit/save behavior, and filtering it when switching agents.
- `git diff --check` passed.

The installed Claude Code 2.1.220 header implementation was also inspected, with an isolated extracted-function check confirming that `0` omits the header and `1` permits it. Backend dependencies were stubbed; this is not an end-to-end Claude/CCR/provider test. Real upstream cache-hit or cost improvements have not been measured.

The earlier PR text claimed verified downstream cache recovery and a universally changing hash. Those claims were not supported by the available evidence and have been corrected.


Yongzhaooo:
Withdrawing this PR after further verification.

The [official Claude Code documentation](https://code.claude.com/docs/en/llm-gateway-protocol#system-prompt-attribution-block) confirms that, since v2.1.181, the attribution block remains stable within a conversation when using a custom base URL. Actual local requests from Claude Code v2.1.220 agree: three successive turns with attribution enabled had identical billing text and system content. Setting the variable to `0` removed the line, but no within-session prefix-stability improvement was demonstrated.

The added regression tests verify the proposed configuration behavior; they do not establish that the profile-wide default change is needed or that it fixes real-provider caching. The original per-turn-hash and verified-cache-recovery claims were too broad, so this PR should not be merged on that basis.

Issue #1791 has been updated with the official version boundary, reproduction results, and limitations. The originally reported OCG cache/usage behavior remains unverified and is not being marked fixed. Thanks for considering the proposal, and apologies for the premature root-cause claim.

PR #1794, opened 2026-09-13
fix(gateway): raise and make configurable the core config-acceptance timeout

### LINKED ISSUE #1761: Gateway fails to start under load: 5s hardcoded config-acceptance timeout measures parent event-loop congestion, not child readiness
## Description

On a moderately loaded host, the managed core gateway (port 3456) fails to come up on every restart/boot, while the main web/management server (3458) works fine. Log line:

```
Failed to start gateway during web startup: Core gateway did not accept runtime config within 5000ms.
```

## Root cause

`packages/core/src/gateway/core-runtime/supervisor.ts` hardcodes `gatewayConfigAcceptanceTimeoutMs = 5_000` and starts the timer when the parent sends `gateway:start`. The child (`gateway-bootstrap.js`) replies `gateway:config-accepted` **after** it finishes `require()`-ing the gateway entry. The timeout therefore measures:

1. child process spawn + boot,
2. the gateway module `require()`,
3. **and how long the parent's event loop takes to get around to reading the IPC reply.**

Point 3 is the killer. During web startup the parent is busy with synchronous init (plugin loading, better-sqlite3 open, marketplace fetch, ...), so the child's ack sits unread in the IPC channel while the 5s timer fires first. The supervisor then tears down a gateway child that actually started fine.

I reproduced this on an idle machine: the same child, spawned standalone with the exact service environment (from `/proc/<pid>/environ`) and the same `gateway:start` IPC message, sends `gateway:config-accepted` in **~450ms**. Yet inside the service the startup races are lost repeatedly: observed 3456 listening at t+7s (after the timer already fired at t+5s), then killed. Success is probabilistic depending on host load — on boot (busiest moment) it essentially always loses.

## Additional problem: no self-healing

When the race is lost, the main process does **not exit** — it just logs the failure and keeps serving 3458. So `Restart=always` (systemd) never triggers, and the gateway stays dead until someone manually restarts the whole service.

## Environment

- v3.0.22, host deployment (`node packages/core/dist/main/server.js --host 0.0.0.0 --port 3458 --no-open`), Ubuntu, Node 22
- 12-core host that runs other CPU/memory-heavy services (python jobs + mongod); no swap

## Suggestions

- Make the acceptance timeout configurable (env/config), e.g. `CCR_GATEWAY_ACCEPT_TIMEOUT_MS`, defaulting to something more forgiving (30-60s), and/or
- Measure only child readiness rather than parent-loop scheduling: treat "child still alive and progressing" differently from "child crashed", and/or
- Retry the gateway startup instead of giving up while the main server keeps running.

Happy to turn the repro probe into a PR if the approach sounds agreeable.

fszcd:
## Summary

`spawnGatewayProcess` starts the config-acceptance timer in the parent when it sends `gateway:start`, but the child only replies `gateway:config-accepted` after it has booted **and** finished `require()`-ing the real gateway entry. The hardcoded 5000ms budget therefore measures child spawn + module load + how congested the parent's event loop is at that moment — not child readiness. On a loaded host (and especially on Windows, where process spawn is slower) the race is lost probabilistically even though the child comes up fine a couple of seconds later, leaving the web UI serving 3458 with a dead gateway on 3456 and no self-healing.

## Changes

- Raise the default config-acceptance budget from 5s to 30s. Genuine child failures still surface immediately via the existing `exit`/`error` handlers, so the larger budget only affects a live-but-slow child.
- Make it tunable via `CCR_GATEWAY_CONFIG_TIMEOUT_MS` (values < 1000ms or unparseable fall back to the default), following the existing `CCR_UPSTREAM_TIMEOUT_MS` precedent.
- The timeout error message now names the env var so users hitting the limit know the knob exists.

Fixes #1761

## Test plan

- `node --test .test-dist/core/test/unit/gateway/gateway-config-acceptance-timeout.test.js`: 4 passed, 0 failed (default, valid override, whitespace/fraction normalization, invalid-value fallback).
- `node --test ".test-dist/core/test/unit/gateway/*.test.js"`: 253 passed, 0 failed, 2 skipped (pre-existing win32 skips).
- `npm run typecheck`: exit 0.
- `npm run test:core` (full): 911 passed / 25 failed on this branch vs 906 passed / 26 failed on unmodified `main` — every failure reproduces identically on the unmodified baseline (Windows environment issues: HOME-sandbox assumptions in profile-service, SQLite usage-store timing, temp-dir `rm` EPERM, etc.). No regressions from this change.


fszcd:
Friendly ping @musistudio — a small, self-contained fix: gateway config-acceptance timeout 5s -> 30s, plus an env override (CCR_GATEWAY_CONFIG_TIMEOUT_MS). Verified locally on Windows. Happy to adjust anything.

PR #1796, opened 2026-09-14
fix(codex): don't inject fast_mode into null config requirements

### LINKED ISSUE #1795: Codex profile (desktop ChatGPT app) fails to launch via CCR: "This app server did not provide application network requirements"
# Codex profile fails to start via CCR: "This app server did not provide application network requirements"

## Environment

- macOS 26.7.0 (arm64), ChatGPT desktop 26.908 (codex-cli bundled `0.154.0-alpha.6.2`)
- CCR latest main (gateway 3456 / management 3459), codex profile `default-codex`
- `codexCliPath` correctly set to the bundled CLI `/Applications/ChatGPT.app/Contents/Resources/codex`

## Steps to reproduce

1. Configure the Codex profile in CCR (model provider = local gateway, `codexCliPath` = bundled codex).
2. Launch the Codex desktop app **through CCR**.
3. App fails immediately after the app-server initialize handshake:

```
error Application network requirements prevented startup
  errorMessage="This app server did not provide application network requirements."
  at St (.../window-all-closed-*.js) at wt.refreshRequirements
```

Launching the **same app directly** (without CCR, `CODEX_HOME=~/.codex`) starts fine.

## What was ruled out (tested)

- **CLI binary**: swapping from standalone `~/.local/bin/codex` (0.147.0) to the bundled codex (0.154.0-alpha.6.2) makes no difference.
- **auth**: shared ChatGPT auth is picked up; `getAuthStatus` correctly returns `{"authMethod":"chatgpt","requiresOpenaiAuth":true}`.
- **sandbox config**: removing `sandbox_mode` / `[sandbox_workspace_write]` changes nothing.

## Root cause

Captured the exact `configRequirements/read` payload in both modes by driving the app-server over stdio:

```jsonc
// DIRECT (CODEX_HOME=~/.codex, bundled codex, no middleware)
{"id":2,"result":{"requirements":null}}

// CCR (middleware + profile CODEX_HOME)
{"id":2,"result":{"requirements":{"featureRequirements":{"fast_mode":true}}}}
```

Direct mode returns `requirements: null`, and the desktop app treats `null` as "no special requirements" and starts normally.

The bug is entirely in the middleware. `configRequirementsRead()` in `packages/core/src/agents/codex/cli-middleware-runtime.ts` runs when the fast-mode shim is enabled and, because the upstream `requirements` is `null`, rebuilds the result into a non-null object containing only `featureRequirements.fast_mode`. The app then sees a non-null `requirements` and expects application network requirements to be present — they are not, so it aborts with "did not provide application network requirements".

## Expected

The middleware must preserve `requirements: null` exactly as the upstream app-server returns it, so the desktop app's startup check passes (same as direct mode).

## Suggested fix

In `configRequirementsRead()`, do not fabricate a requirements object from `null`. Only annotate an already-non-null `requirements` object:

```ts
function configRequirementsRead(existingResult) {
  if (!codexAppFastModeShimEnabled()) return existingResult;
  const result = existingResult && typeof existingResult === "object" && !Array.isArray(existingResult)
    ? { ...existingResult } : {};
  if (result.requirements == null) return existingResult; // preserve null
  const requirements = { ...result.requirements };
  const featureRequirements = /* same guards as current code */ {};
  featureRequirements.fast_mode = true;
  requirements.featureRequirements = featureRequirements;
  result.requirements = requirements;
  return result;
}
```

---

*Verified with manual JSON-RPC over stdio against the bundled codex 0.154.0-alpha.6.2 under both the direct and CCR environments (initialize → configRequirements/read → getAuthStatus).*

Burlesque1:
## Problem

Launching the Codex profile (desktop ChatGPT app) through CCR fails immediately after the app-server initialize handshake:

```
error Application network requirements prevented startup
  errorMessage="This app server did not provide application network requirements."
  at St (.../window-all-closed-*.js) at wt.refreshRequirements
```

Launching the same app directly (without CCR) starts fine.

## Root cause

The bundled app-server returns `requirements: null` when there are no special requirements, and the desktop app treats `null` as "no requirements" and starts normally.

When the fast-mode shim is enabled, `configRequirementsRead()` rebuilt that null result into:

```json
{"requirements": {"featureRequirements": {"fast_mode": true}}}
```

The app sees a **non-null** `requirements` object and expects application network requirements to be present — they are not, so it aborts startup.

Verified over stdio (initialize → `configRequirements/read`):

```jsonc
// DIRECT (no middleware)
{"id":2,"result":{"requirements":null}}

// CCR (middleware + profile CODEX_HOME)
{"id":2,"result":{"requirements":{"featureRequirements":{"fast_mode":true}}}}
```

## Fix

Only annotate an already-non-null `requirements` object; never fabricate one from `null`. When the upstream result's `requirements` is `null`/absent, return the result unchanged.

## Related

- Closes #1795


Burlesque1:
## Verified locally (end-to-end)

Rebuilt the core server with this fix (`npm run build:docker`) and restarted CCR, then launched the Codex desktop app through CCR — it now **starts successfully**, where it previously aborted with "This app server did not provide application network requirements".

Post-fix `configRequirements/read` over stdio:

```jsonc
{"id":2,"result":{"requirements":null}}   // CCR now matches direct mode
```

The middleware also regenerates `ccr-codex-cli-middleware.js` from this fixed script on profile open, so the old (broken) runtime file gets replaced automatically.

musistudio:
I have already fixed it here: https://github.com/musistudio/claude-code-router/commit/16919801701639a7ef0de84e5a68041cf357b541

Burlesque1:
@musistudio thanks — I checked out `16919801701639a7ef0de84e5a68041cf357b541` ("Default Codex app approvals to user review"). It's a solid fix (filling in `application.network` rather than the null-preserving approach here), and I've closed this PR as superseded.

One thing I noticed: that commit doesn't appear to be on `main` yet — `main`'s `cli-middleware-runtime.ts` still has the old `configRequirementsRead` that fabricates `fast_mode` from `null`, so a fresh checkout on `main` still reproduces the "application network requirements" startup failure. The commit also isn't attached to any open PR.

Could you confirm whether it's intended to land on `main`? If it's helpful I can open a PR that carries just that commit so it doesn't get lost.

PR #1802, opened 2026-09-16
fix(docker): Ensure log-body.worker.js is built for Docker deployments

eaglerainbow:
## Problem Description
When running CCR in Docker, the "Log" feature in the management UI fails to load request/response bodies with the error: "**Body formatter worker failed.**"

The browser console shows:

* `NS_ERROR_DOM_NETWORK_ERR` / `NS_ERROR_CORRUPTED_CONTENT` for requests to `/assets/log-body.worker.js`
* MIME type error: "`text/html`" instead of "`application/javascript`" (the worker was being served as the SPA index.html due to nginx fallback)

This occurs because the Docker build process (`build/docker-build.mjs`) was not building the `log-body.worker.js` asset. The worker file is required by the UI's log body formatter (used in `network-logs.tsx` via `createLogBodyFormatterWorker()`) to parse and display request/response payloads in the log viewer.

## Summary of Correction Approach
Added the missing `buildRequestLogBodyWorker` build step to the Docker build pipeline in `build/docker-build.mjs`:

1. Import `buildRequestLogBodyWorker` from the esbuild config
2. Include it in the `Promise.all build` step alongside other renderer assets

This ensures the worker file (`log-body.worker.js`) is compiled and copied to the nginx static assets directory (`/usr/share/nginx/html/assets/`) during Docker image build, making it available for the UI to load with the correct MIME type.

The fix is minimal and follows the existing pattern for building other static assets like `web-client-bridge.js`.


eaglerainbow:
duplicates #1770 
Closing this one in favor of the other.

PR #1811, opened 2026-09-19
feat(profile): add default model list switch with fixed Claude tiers and slot routing

zhi1ong:
## Summary

Adds an opt-in `claudeDefaultModelList` switch to claude-code profiles (UI toggle in the profile advanced settings, next to Enhanced route / routing). When enabled:

- `GET /v1/models` and the Claude CLI bootstrap return a fixed four-tier Claude model list — `claude-fable-5[1m]`, `claude-opus-5[1m]`, `claude-sonnet-5[1m]`, `claude-haiku-4-5` — instead of the allowlist-filtered provider routes, so clients that validate model names see a clean Claude-family surface.
- Requests for these tier names route through the profile model slots (`fableModel`/`opusModel`/`sonnetModel`/`haikuModel`, falling back to the profile default model). These slots were previously only consumed by launcher-mode env injection, so gateway-mode deployments could not select a per-tier target.
- `message_start.model` is rewritten back to the client-requested tier name on the response stream, keeping the client-visible model consistent.

## Implementation notes

- Tier rewriting runs in both the compatibility pipeline and the core runtime router plugin (beforeRouting transform), so the mapping holds in single-runtime and wrapper topologies and on every protocol endpoint (`/v1/messages`, `/v1/chat/completions`, `/v1/responses`).
- The client-requested model is carried to the response stream hook via an internal `x-ccr-client-visible-model` header, which is stripped from untrusted ingress requests like the other `x-ccr-*` route headers.
- The profile allowlist is **extended** with the tier names and slot targets rather than bypassed, so model enforcement stays active for every other model.
- Advertised limits (`max_input_tokens`, capabilities) follow the model the tier actually routes to when it resolves, falling back to tier constants; tiers with no resolvable target are not advertised.
- `count_tokens` maps tier names to the slot target before estimation, and the bootstrap auto-compact windows keep the route-based entries alongside the tier entries.
- Default is **off**: profiles without the switch behave exactly as before. Config accepts `claudeDefaultModelList: true` (snake_case accepted, matching the parse conventions of sibling fields).

## Testing

- New unit tests in `packages/core/test/unit/gateway/claude-default-models.test.mjs`: tier matching (bare/[1m]/case), slot resolution with default fallback, allowlist extension semantics, tier rewrite of requests, claude-code discovered-model rewrite skip, fixed models response shape.
- `tsc --noEmit` clean; the full core suite shows no new failures versus the base commit (the 5 remaining failures reproduce on a clean checkout of the same base).


packages/core/src/gateway/features/model-discovery.ts
```diff
@@ -82,6 +94,18 @@ export function prepareClaudeAppDiscoveredModelRequest(
     return undefined;
   }
 
+  // Default model list: route tier model names through the profile tier slots.
+  if (isClaudeDefaultModelListEnabled(options.profile)) {
```
musistudio: [P1] Apply tier rewriting to the wrapper routing path for every advertised protocol

This mapping is still inside `prepareClaudeAppDiscoveredModelRequest`, which returns before reaching this block unless the path is `/v1/messages`. In `GatewayRequestPipeline`, `/v1/chat/completions` and `/v1/responses` are routed through the internal `/__ccr/route` handler, which calls `routeWithRouter` directly rather than running this core `beforeRouting` transform. At the same time, `createOpenAiCompatibleClaudeDefaultModelsResponse` advertises these tier IDs to clients using those endpoints. In wrapper topology, a request for `claude-sonnet-5[1m]` therefore remains unmapped instead of selecting `sonnetModel`, and can fail model resolution or take the wrong route. Please move the tier rewrite into the protocol-agnostic adapted-body routing path (or the shared internal route handler) and cover both endpoints with integration tests.

packages/core/src/gateway/features/model-discovery.ts
```diff
@@ -245,6 +295,128 @@ function createOpenAICompatibleGatewayModelsResponse(config: AppConfig, profile?
...
+      type: "model"
+    }];
+  });
+}
+
+
+function createClaudeDefaultTierAutoCompactWindows(): Record<string, number> {
+  const windows: Record<string, number> = {};
+  for (const tier of claudeDefaultModelTiers) {
+    const compactWindow = claudeCliAutoCompactWindow(tier.maxInputTokens);
```
musistudio: [P2] Derive the auto-compact window from the resolved slot target

Using `tier.maxInputTokens` makes the first three tier names unconditionally receive a 1,000,000-token compact window, even when the selected profile slot points to a 120k/200k model. `claudeDefaultTierEntry` correctly advertises the target model’s actual `max_input_tokens`, so the bootstrap payload can now contradict model discovery and Claude may compact only after the upstream context limit has already been exceeded. Please pass `config`/`profile` here and compute each window from the resolved tier entry’s `maxInputTokens`; a regression test with provider metadata such as `contextWindow: 120000` would expose the mismatch.

musistudio (commented):
I found two functional issues in the opt-in default-model-list path. The focused tests pass, but they do not exercise the wrapper routing topology or target-aware auto-compaction described in the inline comments.

zhi1ong:
Thanks for the review — both points are addressed in 9d4b946.

**P1 (wrapper topology, non-`/v1/messages` endpoints):** the tier rewrite now lives in a protocol-agnostic pipeline step (`prepareClaudeDefaultTierModelRequest`) that runs before routing for `/v1/messages`, `/v1/chat/completions` and `/v1/responses`, so wrapper-topology requests for advertised tier ids map to the profile slots. The core runtime router plugin keeps applying the same mapping through its beforeRouting transform in single-runtime topology, and the `/v1/messages`-only discovered-model prepper is back to its upstream shape. A new integration test (`default-model-list-tier-routing-http.test.mjs`) drives all three endpoints through the compatibility pipeline against a mock core and asserts the rewritten request models.

**P2 (auto-compact windows):** the bootstrap windows are now derived per tier from the resolved slot target's advertised `max_input_tokens` (same `claudeDefaultTierEntry` helper model discovery uses), with a regression test using a `contextWindow: 120000` slot.

`tsc --noEmit`, the docker build and the full core suite show no new failures versus the base commit.

packages/core/src/gateway/features/model-discovery.ts
zhi1ong: Fixed in 9d4b946: the tier rewrite moved into the protocol-agnostic pipeline step `prepareClaudeDefaultTierModelRequest`, which runs before routing on `/v1/messages`, `/v1/chat/completions` and `/v1/responses` in wrapper topology (the core runtime router plugin still applies the same mapping via its `beforeRouting` transform in single-runtime topology, and the `/v1/messages`-only discovered-model prepper is back to upstream shape). Covered by `packages/core/test/integration/gateway/default-model-list-tier-routing-http.test.mjs`, which drives all three endpoints through the compatibility pipeline against a mock core and asserts the rewritten request models.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/gateway/features/model-discovery.ts
zhi1ong: Fixed in 9d4b946: `createClaudeDefaultTierAutoCompactWindows` now takes `config`/`profile` and derives each window from the resolved tier entry's `maxInputTokens` (`claudeDefaultTierEntry`, the same helper model discovery uses), so bootstrap agrees with the advertised `max_input_tokens`. Added a regression test with a `contextWindow: 120000` slot asserting the tier windows follow the target.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

zhi1ong:
Found a gap while testing: the client-visible model rewrite was SSE-only, so non-streaming responses kept the upstream model name. New commit 131b527 extends it to non-streaming Anthropic JSON in both runtimes (pipeline transform + engine response hook), with tests.

